PR #57, opened 2015-02-24
only add content-type header for PUT and POST requests

redonkulus:
@lingyan @Vijar


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

lingyan:
:ship:

PR #68, opened 2015-04-01
Use JSON.parse on query string values in case they are arrays or objects

Vijar:
Also, use supertest for better coverage.

@mridgway @redonkulus

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -6,18 +6,34 @@
 /**
  * list of registered fetchers
  */
-var OP_READ = 'read',
-    OP_CREATE = 'create',
-    OP_UPDATE = 'update',
-    GET = 'GET',
-    qs = require('querystring');
+var OP_READ = 'read';
+var OP_CREATE = 'create';
+var OP_UPDATE = 'update';
+var GET = 'GET';
+var qs = require('querystring');
+var debug = require('debug')('Fetchr');
+
+function parseValue(value) {
+    // take care of value of type: array, object
+    try {
+        return JSON.parse(value);
+    } catch (e) {
+        return value;
+    }
+}
+
+function parseParamValues (params) {
+    return Object.keys(params).reduce(function (parsed, curr) {
+        parsed[curr] = parseValue(params[curr]);
+        return parsed;
+    }, {});
+}
 
 /*
  * @module createFetcherClass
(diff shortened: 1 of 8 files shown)


yahoocla:
CLA is valid!

libs/fetcher.js
```diff
@@ -6,18 +6,34 @@
...
+var OP_UPDATE = 'update';
+var GET = 'GET';
+var qs = require('querystring');
+var debug = require('debug')('Fetchr');
+
+function parseValue(value) {
+    // take care of value of type: array, object
+    try {
+        return JSON.parse(value);
+    } catch (ignore) {
```
jedireza: Rename `ignore` to `e`, `error` or `exception`?

jedireza:
:+1:

package.json
```diff
@@ -27,13 +27,16 @@
     "xhr": "^2.0.0"
   },
   "devDependencies": {
+    "body-parser": "~1.12.2",
```
redonkulus: ^

package.json
```diff
@@ -27,13 +27,16 @@
...
     "chai": "^2.0.0",
     "coveralls": "^2.11.1",
+    "express": "~4.12.3",
```
redonkulus: ^

libs/fetcher.js
Vijar: JSLint used to expect this to be `ignore` but I think it can be anything now.

Vijar:
@redonkulus @jedireza updated.

tests/unit/libs/fetcher.client.js
```diff
@@ -39,42 +44,29 @@ describe('Client Fetcher', function () {
...
-                    pathname = decodeURIComponent(pathname);
-                    pathname = pathname.split(';');
-                    expect(pathname.shift()).to.equal('/api/resource/' + resource);
-                    while(!!(pair = pathname.shift())) {
-                        var pair = pair.split('=');
-
-                        var k = pair[0],
-                            v = pair[1];
-
-                        //hacky because of array in querystring, but its fine
```
Vijar: @redonkulus was the hack to get around arrays not being serialized properly before...

redonkulus:
:shipit:

PR #65, opened 2015-03-30
Make API path resource configurable via pathPrefix

lenage:
In this PR, 

I added an global var as default pathPrefix(/resource/), and provide an option `this.pathPrefix` to make it configurable.

NOTE: if you want to remove `/resource/` prefix in path, please set `this.pathPrefix` to `''`.

PR #69, opened 2015-04-07
Remove path prefix '/resource' for endpoint url

lenage:
Remove "/resource" prefix in url.

see #65  for more details


yahoocla:
CLA is valid!

jedireza:
:+1: good stuff

PR #70, opened 2015-04-10
support custom response headers

lingyan:
@Vijar @mridgway 

It's common for data fetchers to want to set custom response headers, especially headers like cache-control for GET requests.

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -106,6 +106,9 @@ function parseParamValues (params) {
                             return;
                         }
                         meta = meta || {};
+                        if (meta.headers) {
+                            res.set(meta.headers);
+                        }
                         res.status(meta.statusCode || 200).json(data);
                     }
                 };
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

redonkulus:
:shipit:

jedireza:
:+1:

Vijar:
@lingyan @redonkulus 

Now that fetchr does this directly, do we still need this: https://github.com/yahoo/fluxible-plugin-fetchr/blob/master/lib/fetchr-plugin.js#L27 ?

lingyan:
@Vijar  That's still needed.  Because this PR only applies to xhr endpoint, when you use the fetchr middleware.   `getServiceMeta()` in fetchr-plugin is used for page request to combine headers from service calls.

Vijar:
Ah, you're right!

PR #74, opened 2015-05-05
Allow JSON thru on error

dsteinbach:
@mridgway @lingyan @redonkulus 

This change allows JSON from error responses to be passed back to our app but I am not certain if there is a better way of achieving this. We are receiving error codes and messages from APIs that we would like to run through i18n and display to our users.  

We could possibly pass statusCode and a new property called `jsonMessage` from the context.service callback which Fetchr might handle differently? 
```
callback({
    statusCode: err.statusCode,
    jsonMessage: err
}, res);
```

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -175,7 +175,11 @@ function io(url, options) {
                 errMessage = status ? 'Error ' + status : 'Internal XMLHttpRequest Error';
             }
 
-            err = new Error(errMessage);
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

dsteinbach:
This is the wrong solution. I should just be passing response argument even if there is an error.

PR #75, opened 2015-05-05
Fix install instructions

guyellis:


yahoocla:
CLA is valid!

examples/simple/README.md
```diff
@@ -4,8 +4,12 @@
 
 ### Install
 
+Start fetchr root:
+
 ```js
 npm install
+cd examples/simple
+npm install
```
redonkulus: there are 2 npm installs here, I think thats a typo. also, the instructions are assuming the user is in the examples/simple folder already (since the readme is there). I'm not so sure its needed.

guyellis: @redonkulus If you remove the node_modules folder in the root and then follow the original instructions you'll see that they fail. i.e. the examples depend on `npm i` having been run in the root *and* in the examples/simple folder before it will work.

guyellis: The app.js references libs/fetcher.client.js in the root which requires lodash which is not a dependency in the examples.

redonkulus: Got it. ok makes sense.

PR #76, opened 2015-05-06
Pass response body for errors

dsteinbach:
We do not pass response in the event of errors but often APIs provide lots of good details in an error response, such as error codes that we can use to notify the user about. This change attaches body of api response to the err object.


yahoocla:
CLA is valid!

Vijar:
LGTM.

@dsteinbach can you squash into one commit?

PR #80, opened 2015-05-21
Return json for error response

Vijar:
@lingyan 

This is so we're consistent in sending the same format response from the middleware.


yahoocla:
CLA is valid!

lingyan:
:+1:

Issue #83, opened 2015-05-28
Add X-Requested-With: XMLHttpRequest header


lingyan:
It is a common header used for identify Ajax requests: http://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields

Vijar:
Fixed by #84

neilius:
What are your thoughts on disabling (or allowing a configuration value that disables) the X-Requested-With header for CORS-based implementations? When performing CORS requests, the X-Requested-With forces a pre-flight OPTIONS request that would be nice to eliminate. jQuery, for example, [is structured this way](https://github.com/jquery/jquery/blob/master/src/ajax/xhr.js#L60)

lingyan:
@neilius That sounds good.  We can turn off X-Requested-With header for cross domain requests.  Do you want to file a PR?  :)

neilius:
Yeah, certainly

lingyan:
Thanks!

PR #84, opened 2015-06-01
add X-Requested-With header; bug fix

lingyan:
@Vijar 

* add X-Requested-With header (addresses https://github.com/yahoo/fetchr/issues/83)
* fix a bug that broke retry, discovered while adding unit test for http.client.js
* had to turn off retry by default to keep backward compatibility

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -18,7 +18,7 @@ var _ = require('lodash'),
         timeout: 3000,
         retry: {
             interval: 200,
-            max_retries: 2
+            max_retries: 0
         }
     },
     CONTENT_TYPE = 'Content-Type',
@@ -39,7 +39,9 @@ if (!String.prototype.trim) {
 }
 
 function normalizeHeaders(headers, method) {
-    var normalized = {};
+    var normalized = {
+        'X-Requested-With': 'XMLHttpRequest'
+    };
     var needContentType = (method === METHOD_PUT || method === METHOD_POST);
     _.forEach(headers, function (v, field) {
         if (field.toLowerCase() === 'content-type') {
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

libs/util/http.client.js
```diff
@@ -18,7 +18,7 @@ var _ = require('lodash'),
         timeout: 3000,
         retry: {
             interval: 200,
-            max_retries: 2
+            max_retries: 0
```
lingyan: Turning retry off to keep backward compatible behavior of no automatic retry, which was due to a bug that is fixed in this commit.

libs/util/http.client.js
```diff
@@ -139,7 +141,7 @@ function doXhr(method, url, headers, data, config, callback) {
                 callback(NULL, response);
             },
             failure : function (err, response) {
-                if (!shouldRetry(method, config, response.status)) {
+                if (!shouldRetry(method, config, response.statusCode)) {
```
lingyan: Fix the retry bug

redonkulus:
:+1:

libs/util/http.client.js
```diff
@@ -39,7 +39,9 @@ if (!String.prototype.trim) {
 }
 
 function normalizeHeaders(headers, method) {
-    var normalized = {};
+    var normalized = {
+        'X-Requested-With': 'XMLHttpRequest'
+    };
```
lingyan: `X-Requested-With` is a common header used for identify Ajax requests: http://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields

Vijar:
:shipit:

PR #85, opened 2015-06-01
move default constructGetUri function to its own file

lingyan:
@Vijar @redonkulus

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -13,27 +13,27 @@
  * batches requests into one request.
  * @module Fetcher
  */
-var REST = require('./util/http.client'),
-    debug = require('debug')('FetchrClient'),
-    lodash = {
+var REST = require('./util/http.client');
+var debug = require('debug')('FetchrClient');
+var lodash = {
         isArray: require('lodash/lang/isArray'),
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

libs/fetcher.client.js
```diff
@@ -382,13 +345,8 @@ Queue.prototype = {
...
             if (!use_post) {
-                if (lodash.isFunction(config.constructGetUri)) {
-                    get_uri = config.constructGetUri.call(this, uri, request.resource, request.params, config);
-                }
-
-                if (!get_uri) {
-                    get_uri = this._defaultConstructGetUri(uri, request.resource, request.params, config);
-                }
+                var getUriFn = lodash.isFunction(config.constructGetUri) ? config.constructGetUri : defaultConstructGetUri;
+                var get_uri = getUriFn.call(this, uri, request.resource, request.params, config, this.context);
```
lingyan: I did add one more param `this.context` to the construct get uri function, since that is most likely the only thing the function needs from `this` object (at least for the default `constructGetUri` function).  Should not break compatibility.

Vijar: I like that it is explicit.

Vijar:
:shipit:

Issue #87, opened 2015-06-02
Remove `config` param from service


Vijar:
Right now `config` does not work the same on client and server. On the client it controls things like xhr timeouts and is never serialized and sent to the fetchr middleware, on the server it is just data that is passed into the service.

This is inconsistent. We can assume not many people use it because it took so long to find. Since config has been optional for a while now, we should just remove it and bump minor.

@redonkulus  @lingyan

redonkulus:
what are the changes to the interface? how does this affect users? are we just combining params and configs? if xhr is still used on the client, how does that get used in the new API?

escalona:
I've run into problems with this when calling a service from the client where I'm expecting config to be defined.

Instead of removing the config object, could it simply be renamed in fetcher.client to xhrConfig with updated docs to reflect the difference?

Vijar:
@escalona I'm leaning towards this approach too.

escalona:
@Vijar I've created a PR with an updated config name as well as additional docs. Decided to use the name "clientConfig" as it appears that object is responsible for more than just XHR config options.

I'd like to document the other options as well but need more insight into a few of them. Will create a separate issue to track.

Vijar:
Fixed by #92

PR #92, opened 2015-06-09
Fetcher.client config object name

escalona:
In response to: #87 
- Update fetcher.client config name to clientConfig
- Add docs to reflect difference between client and server side config objects.


yahoocla:
CLA is valid!

Vijar:
:ship:

Vijar:
Resolves #32

Issue #94, opened 2015-06-09
Fetcher.middleware config undefined


escalona:
The config object is never serialized to fetchr middleware when a request is made through the client. This causes inconsistent behavior within a service that is called from both the client and the server.

On all requests other than "GET" requests, config always returns undefined in the service. Proposed fix is to have a fallback config object returned at:
https://github.com/yahoo/fetchr/blob/e00ceadabcf2bc55a0016cbc0cb6cf36740096f2/libs/fetcher.js#L126-L133

Similar issue #87

escalona:
Fixed by #95

PR #95, opened 2015-06-09
Add fallback config object

escalona:
For requests other than "GET" requests, add a fallback config object so config is never undefined when service is called.

Is response to: #94

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -130,7 +130,7 @@ function parseParamValues (params) {
                     operation: singleRequest.operation,
                     params: singleRequest.params,
                     body: singleRequest.body || {},
-                    config: singleRequest.config,
+                    config: {},
                     callback: function(err, data, meta) {
                         meta = meta || {};
                         if (meta.headers) {


yahoocla:
CLA is valid!

Vijar:
@escalona looking through the code ([this](https://github.com/escalona/fetchr/blob/130a37f6804d18bc28e4cf678335d6f11381c94d/libs/fetcher.client.js#L369) and [this](https://github.com/escalona/fetchr/blob/130a37f6804d18bc28e4cf678335d6f11381c94d/libs/fetcher.client.js#L28)), I think config is always empty in the middleware for both GET and POST.

So I would be in favor of replacing `singleRequest.config` with an empty object.

escalona:
Had a hunch that was the case. Will make that change

Vijar:
:+1: 

@escalona can you squash into one commit? I'll merge after.

escalona:
@Vijar Thanks, commits have been squashed

Vijar:
Thanks!

mdressman:
:+1:

PR #98, opened 2015-06-12
handle error 400 json responses and standard string responses

### LINKED ISSUE #90: Fetcher client not handling error properly
Recently a change was made so that the errors are now being sent as a JSON object instead of a pure string https://github.com/yahoo/fetchr/commit/deef81c24055094ac225848f03898dd04b4abd1c. However, this change has not been reflected properly in the fetcher client and is still expecting it as a string.

The actual result of this [for the callback method] is that the error parameter is now an error object (which is the same as before i believe), with the message property being the JSON string that got sent from the server rather than just the message string itself.

If the intention was to make it that anyone from the server side could pass more than just a message, then perhaps the response parameter should be a parsed JSON object instead of being undefined?

In regards to where the work should be done,  I'm not sure if it's more appropriate to update the http client instead to cover all errors in one go, or if it should be parsed in each of the REST methods inside fetcher client.

I could look at making the changes, but what was the original intent of the commit that caused this issue?

davidhouweling:
This should hopefully fix #90. I've tried to incorporate appropriate tests based on scenarios that I've seen before the commit created the problem, as well as scenarios I saw from an IIS server housing node where it stripped out the body content.

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -168,18 +168,27 @@ function io(url, options) {
         useXDR: options.cors
     }, function (err, resp, body) {
         var status = resp.statusCode;
-        var errMessage;
+        var errMessage, errBody;
 
         if (!err && (status === 0 || (status >= 400 && status < 600))) {
             if (typeof body === 'string') {
-                errMessage = body;
+                try {
+                    errBody = JSON.parse(body);
+                    if (errBody.message) {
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

Vijar:
:ship:

PR #102, opened 2015-06-17
[Fixes #101] isRegistered handles sub resources

### LINKED ISSUE #101: Bad resource
Hi,

On new version, you have add this
```
if (!Fetcher.isRegistered(resource)) {
  error = fumble.http.badRequest('Invalid Fetchr Access', {
    debug: 'Bad resource ' + resource
  });
  error.source = 'fetchr';
  return next(error);
}
```
But... what's happening when i want use "todo.toggleAll" on resource name (https://github.com/yahoo/flux-examples/blob/master/todo/actions/toggleAll.js#L11, https://github.com/yahoo/flux-examples/blob/master/todo/services/todo.js#L43), it's just check resource var, when it should be
```
resource.split('.')[0]
```

After maybe i completely wrong (suddenly i do not really know what to do).

Thanks !

Vijar:
@redonkulus

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -75,10 +75,11 @@ function parseParamValues (params) {
      */
     Fetcher.getFetcher = function (name) {
         //Access fetcher by name
-        if (!name || !Fetcher.fetchers[name]) {
+        var fetcher = Fetcher.isRegistered(name);
+        if (!fetcher) {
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #103, opened 2015-06-17
Remove X-Requested-With header when performing CORS requests

neilius:
This PR prevents the `X-Requested-With` header from being set when making CORS requests, since that header causes an undesired preflight OPTIONS request. See #83 for original mention.

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -43,10 +43,11 @@ if (!String.prototype.trim) {
   };
 }
 
-function normalizeHeaders(headers, method) {
-    var normalized = {
-        'X-Requested-With': 'XMLHttpRequest'
-    };
+function normalizeHeaders(headers, method, isCors) {
+    var normalized = {};
+    if (!isCors) {
+        normalized['X-Requested-With'] = 'XMLHttpRequest';
+    }
     var needContentType = (method === METHOD_PUT || method === METHOD_POST);
     _.forEach(headers, function (v, field) {
         if (field.toLowerCase() === 'content-type') {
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

libs/util/http.client.js
```diff
@@ -43,10 +43,11 @@ if (!String.prototype.trim) {
   };
 }
 
-function normalizeHeaders(headers, method) {
-    var normalized = {
-        'X-Requested-With': 'XMLHttpRequest'
-    };
+function normalizeHeaders(headers, method, isCors) {
+    var normalized = {};
+    if(!isCors) {
```
redonkulus: add space after if

libs/util/http.client.js
```diff
@@ -130,8 +131,8 @@ function mergeConfig(config) {
 function doXhr(method, url, headers, data, config, callback) {
     var options, timeout;
 
+    headers = normalizeHeaders(headers, method, config.cors);
```
redonkulus: why was this moved up? do we not want it merged first?

neilius: `mergeConfig` doesn't appear to actually merge the config passed in, it just returns a new config object using some of the properties from the passed-in config to initialize the new one. This means that the `cors` property is no longer present when `normalizeHeaders` is called. I moved it up so that I could access the `cors` property before `mergeConfig` wipes it out. Perhaps `mergeConfig` needs to be reworked a bit?

redonkulus: got it.

Vijar:
:ship: 

@neilius can you squash your commits into one?

neilius:
@Vijar sure thing, it's now squashed

Vijar:
Thanks!

Issue #58, opened 2015-03-12
Client-side timeouts


gpbl:
Client-side requests often get answers after the [3 seconds timeout](https://github.com/yahoo/fetchr/blob/master/libs/util/http.client.js#L19-L21): i.e. mobile users with bad connections or requests to external services.

It would be useful being able to set the timeout globally, instead of passing it as config to each service. For example, I'd put it to 30s, after which it seems reasonable to show an error. The default is imho too short.

What do you think?

koulmomo:
I like the idea of being able to change the default configs all at once, but I think the default should still be 3 seconds.

I might be wrong, but I *feel* that in most cases (especially in development) it would be helpful to identify which of your services are slow so that you can either fine tune them, or adjust timeouts for those specific services.

gpbl:
@koulmomo does the timeout affect users with slow connections? I have reports of users connecting from mobile. I've also seen it on my computer when it was heavy downloading stuff. In this case the problem is not the service, but the client – hence my need to set the timeout for *all* the service calls.

koulmomo:
@gpbl I'm saying that you **SHOULD** be able to change the defaults for services **ALL AT ONCE**, but if the developer does not choose to do so, it should still default to 3s.

```js
// illustrating a point, not the proposed api
var fetcher = require('fetchr'); // fetcher defaults to 3s timeout
fetcher.defaults({
    timeout: 30 * 1000
}); // fetcher defaults to 30s timeouts
```

geekyme:
@koulmomo I think that @gpbl is right about the 3s default. It does not appear to make sense. And before I stumbled onto this thread, I didn't know about the 3s and spent a few hours debugging this. I build web apps for mobile access and these networks are often unpredictable to the point that 3s timeouts would not be sufficient.

redonkulus:
Fixed by https://github.com/yahoo/fetchr/pull/107

PR #107, opened 2015-07-04
Allow setting of xhr timeout for all request

escalona:
In response to #58

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -28,6 +28,7 @@ var lodash = {
 var CORE_REQUEST_FIELDS = ['resource', 'operation', 'params', 'body'];
 var DEFAULT_GUID = 'g0';
 var DEFAULT_XHR_PATH = '/api';
+var DEFAULT_XHR_TIMEOUT = 3000;
 // By default, wait for 20ms to trigger sweep of the queue, after an item is added to the queue.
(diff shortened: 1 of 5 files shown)


escalona:
If this gets accepted I'll create a PR for fluxible-plugin-fetchr to allow setting this option

libs/fetcher.client.js
```diff
@@ -146,6 +147,7 @@ Queue.prototype = {
      * @class FetcherClient
      * @param {object} options configuration options for Fetcher
      * @param {string} [options.xhrPath="/api"] The path for XHR requests
+     * @param {number} [options.xhrTimout=3000] Timeout for all XHR requests
```
Vijar: Perhaps `Timeout in milliseconds...` to specify the units

Vijar:
@escalona 
We should also get rid of `DEFAULT_CONFIG.timeout`([here](https://github.com/yahoo/fetchr/blob/master/libs/util/http.client.js#L93) and [here](https://github.com/yahoo/fetchr/blob/master/libs/util/http.client.js#L23)) now that we always fallback to the `xhrTimeout` value.

escalona:
@Vijar Good catch. Updated http.client

Vijar:
:ship:

Issue #97, opened 2015-06-12
return data is lost on error


lukemadera:
Following the standard node convention, a callback with a non-null first argument signifies an error and often there is valuable error information passed back. However, in the current implementation, on error, the data is lost (overwritten with `message: err.message || 'request failed'`).

https://github.com/yahoo/fetchr/blob/master/libs/fetcher.js#L109

This is a significant issue since we are currently trying to validate form input fields and need a key value error object of errors per input. More generally, whatever error information or data is supplied should be passed through.
Any particular reason neither error nor data are passed through?

Solution should be pretty simple, such as:

```js
if (err) {
  res.status(err.statusCode || 400).json(err);
  return;
}
```

davidhouweling:
After writing up my fix for #90 I understand now why they did what they did initially. I think a more appropriate solution would be to start with a generic error object with a message attribute, and then extend it with the data your wish to have transmitted across the wire. This way the original added value isn't lost.

lukemadera:
You mean like this @davidhouweling ?

```js
if (err) {
  err =_.extend({message: err.message || 'request failed'}, err);
  res.status(err.statusCode || 400).json(err);
  return;
}
```

davidhouweling:
Yes something along those lines, though I wouldn't pull in the whole underscore library to do it, even though it is just execution on server side (from memory).

I'd probably go with Object.assign if you can get away with it, or an npm module that implements it. This way when the codebase does move entirely to ES6, the code itself is already there and not requiring refactoring.

lukemadera:
Yeah that was just a proof of concept but I agree @davidhouweling  - how's this:

if (err) {
  if(!err.message) {
    err.message ='request failed';
  }
  res.status(err.statusCode || 400).json(err);
  return;
}

Vijar:
@lukemadera 

I think we could support this.

I was just experimenting with the code sample you have and it looks like passing in an error object into `res.json` has weird behavior. I've noticed that it ignores the `message` property.

lukemadera:
Great @Vijar ! What seems to be the problem with message? It just gets omitted?

Vijar:
Yeah it just doesn't get serialized.

jeffkole:
Yes, the fields of an `Error` are not enumerable, so they are not serialized as JSON:

`Object.keys(new Error('request failed'))` results in `[]`

What about if we did this:

```javascript
if (err) {
  var data = {
    message: err.message || 'request failed',
    name: err.name,
[...]

PR #113, opened 2015-08-05
Chatty errors

KATT:
As I suggested in https://github.com/yahoo/fetchr/issues/97#issuecomment-127982042

Possibility to return more data in errors, by supplying a method called `.toSafeJSON()` on the error object itself.

Thoughts? Suggestions on another name for the method?

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -28,6 +28,31 @@ function parseParamValues (params) {
     }, {});
 }
 
+/**
+ * Takes an error and resolves output and statusCode to respond to client with
+ *
+ * @param  {Error} JavaScript error
+ * @return {Object} object with resolved statusCode & output
+ */
+function getErrorResponse(err) {
+    var statusCode = err.statusCode || 400;
+    var output = {
+        message: 'request failed'
+    };
+
+    if (typeof err.output !== 'undefined') {
+        output = err.output;
+    } else if (err.message) {
+        output.message = err.message;
+    }
+
+
+    return {
+        statusCode: statusCode,
+        output: output
+    };
+}
+
 
 /**
  * A Request instance represents a single fetcher request.
@@ -68,7 +93,7 @@ Request.prototype.params = function (params) {
  * Add body to this fetcher request
  * @method body
  * @memberof Request
- * @param {Object} body The JSON object that contains the resource data being updated for this request. 
+ * @param {Object} body The JSON object that contains the resource data being updated for this request.
  *                      Not used for read and delete operations.
  * @chainable
  */
@@ -230,9 +255,8 @@ Fetcher.middleware = function () {
                         res.set(meta.headers);
                     }
                     if (err) {
-                        res.status(err.statusCode || 400).json({
-                            message: err.message || 'request failed'
-                        });
+                        var errResponse = getErrorResponse(err);
+                        res.status(errResponse.statusCode).json(errResponse.output);
                         return;
                     }
                     res.status(meta.statusCode || 200).json(data);
@@ -269,9 +293,8 @@ Fetcher.middleware = function () {
                         res.set(meta.headers);
                     }
                     if(err) {
-                        res.status(err.statusCode || 400).json({
-                            message: err.message || 'request failed'
-                        });
+                        var errResponse = getErrorResponse(err);
+                        res.status(errResponse.statusCode).json(errResponse.output);
                         return;
                     }
                     var responseObj = {};
--- a/tests/unit/libs/fetcher.js
+++ b/tests/unit/libs/fetcher.js
@@ -57,7 +57,7 @@ describe('Server Fetcher', function () {
         expect(getService).to.not.throw;
         expect(getService()).to.deep.equal(mockService);
     });
-    
+
     it('should be able to update options for the fetchr instance', function () {
         fetcher.updateOptions({req: {foo: 'bar'}});
         expect(fetcher.options.req.foo).to.equal('bar');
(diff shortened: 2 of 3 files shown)


Vijar:
@KATT Thanks for doing a PR!

I propose that we don't need the overhead of a `.toSafeJSON()` method. I don't see a use case for an extra function call. I believe it can just be a simple object.

Ideally, usage would look like this:
```js
// in your data servide
create: function (resource, params, body, config, callback) {
    if ( body.username === '' ) {
        var err = new Error('Validation Error');
        err.meta = {
            username: 'Username field is required!'
        }
        callback(err);
    }
}
```

Fetchr middleware should just combine `err.meta` object and `err.message`, then serialize and send to client.

KATT:
@Vijar 

It's more flexible if you let the user return whatever object you want. You might not want or you might need it not to be all nested under `.meta`.

If we were go along this route I'd prefer if we could override what it returned with a property like `.output`.

#### in your data service

```js
create: function (resource, params, body, config, callback) {
	var validationErrors = [];
    if ( body.username === '' ) {
    	validationErrors.push({
    		name: 'username',
    		message: 'Username field required',
    	})
    }

    if (validationErrors.length > 0) {
	    var err = new Error('Validation Error');
	    err.statusCode = 400;
	    err.output = {
	        validationErrors: validationErrors,
	        message: 'Validation failed',
	    }
	    callback(err);
	    return;
	}

	// [..]
}
```

#### Outputting 

**statusCode:** 400

**response:** :point_down: 

```json
{
  "validationErrors": [
    {
      "name": "username",
      "message": "Username field required!"
    }
  ],
  "message": "Validation failed"
}
```

Vijar:
Okay, I'm just trying to avoid a breaking change so I don't think we should
ignore the message property.

How about if we fallback to the message property if the user does not set
the output property?

KATT:
Yeah, of course! 

None of my suggestion (or this PR) are breaking changes. If you look [here](https://github.com/KATT/fetchr/blob/feature/chatty-error/libs/fetcher.js#L130-L139), if you don't have the `.toSafeJSON()` method it'll not change how fetchr behaves. I was thinking the same with `.output`.

KATT:
@Vijar happier with this? using an arbitrary `output` property instead. I can squash it all if you want.

tests/unit/libs/fetcher.js
```diff
@@ -350,10 +366,25 @@ describe('Server Fetcher', function () {
                 return function(done) {
                     var operation = 'read',
                         statusCodeSet = false,
+
                         req = {
-                            method: 'GET',
-                            path: '/' + mockErrorService.name + ';' + qs.stringify(params, ';')
+                            method: 'POST',
```
Vijar: Should be GET right?

libs/fetcher.js
```diff
@@ -269,9 +288,9 @@ Fetcher.middleware = function () {
                         res.set(meta.headers);
                     }
                     if(err) {
-                        res.status(err.statusCode || 400).json({
-                            message: err.message || 'request failed'
-                        });
+                        var errResponse = getErrorResponse(err);
+
```
Vijar: We can do away with the empty line.

libs/fetcher.js
```diff
@@ -230,9 +249,9 @@ Fetcher.middleware = function () {
                         res.set(meta.headers);
                     }
                     if (err) {
-                        res.status(err.statusCode || 400).json({
-                            message: err.message || 'request failed'
-                        });
+                        var errResponse = getErrorResponse(err);
+
```
Vijar: We can do away with the empty line.

Vijar:
@KATT left some comments, but looks good otherwise. Can you look into why the test coverage dropped? Maybe the GET/POST comment I made will fix it.

tests/unit/libs/fetcher.js
KATT: good catch!

KATT:
@Vijar: okay, it should be ready to merge now. Again, let me know if you want me to squash it :)

libs/fetcher.js
```diff
@@ -28,6 +28,25 @@ function parseParamValues (params) {
     }, {});
 }
 
+function getErrorResponse(err) {
```
redonkulus: Can you add a docblock to this function that describes each property excepted for this function?

libs/fetcher.js
```diff
@@ -28,6 +28,25 @@ function parseParamValues (params) {
...
+    var statusCode = err.statusCode || 400;
+    var output = {
+        message: 'request failed'
+    };
+
+    if (typeof err.output !== 'undefined') {
+        output = err.output;
+    } else if (err.message) {
```
redonkulus: is this for backwards compatibility? could we just remove the `else` and make the `if (err.message)` its own block?

KATT: > is this for backwards compatibility?

Yes.

> could we just remove the else and make the `if (err.message)` its own block?


No, that would tamper with your special `.output` if you have thrown an error with a message, which would lead to some confusing outputs & defeats the purpose of being able to customise the output.

redonkulus: Got it 😀

redonkulus:
One last thing we should do is document this in the readme so others know how to use it. 

Other than that it looks good to me.

redonkulus:
And squash of course.

Vijar:
:ship:

redonkulus:
@KATT I merged this to get the feature in. Please feel free to open another PR for the documentation update. Thanks for contributing!

PR #120, opened 2015-08-28
[Fix #119] Don't cast big integers to numbers, leave as strings.

### LINKED ISSUE #119: Numeric value is cast to the number when `parseParamValues` in middleware
My problem is the numeric value is cast to the number would be.
It is occurring in GET request when the parsing by the middleware.

for example... http://example.com/fetchr/Example;id_str=9007199254740992000000

```javascript
function read(req, src, params, options, callback) {
  console.log(params.id_str); // => 9.007199254740992e+21
}
```

What a better solution?
thanks

Vijar:
@redonkulus

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -15,7 +15,16 @@ var objectAssign = require('object-assign');
 function parseValue(value) {
     // take care of value of type: array, object
     try {
-        return JSON.parse(value);
+        var ret = JSON.parse(value);
+        // Numbers larger than MAX_SAFE_INTEGER will contain rounding errors so,
+        // we will just leave them as strings instead. The length > 15 check is because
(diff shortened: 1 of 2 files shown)


Vijar:
The `length > 15` check is because the[ `MAX_SAFE_INTEGER` in javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) is `9007199254740991` which has a length of 16.

Using a `length > 15` check is still safe and faster than trying to use a library supporting big integers.

redonkulus:
Code looks good, lets have @mridgway do a once over too.

Also would be good to put your comment above into the code for future reference.

mridgway:
Looks good to me. A little weird, but I understand the problem.

PR #121, opened 2015-09-01
add context picker support

kaesonho:
@Vijar @redonkulus @mridgway @lingyan 

Add a new option `contextPicker`, so that users can define how to pick the context values, for example they can ignore certain key for certain method.

the pick function is exactly the predicate function of `lodash/object/pick`, 

```js
var fetcher = new Fetcher({
    context: { // These context values are persisted with XHR calls as query params
        _csrf: 'Ax89D94j',
        device: 'desktop'
    },
    contextPicker: {
        GET: function (value, key, object) {
            // for example, if you don't enable CSRF protection for GET, you are able to ignore it with the url
            if (key === '_csrf') {
                return false;
            }
            rerurn true;
        }
        // for other method e.g., POST, if you don't define the picker, it will pick the entire context object
    }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,39 @@ fetcher
 
 For requests from the server, the config object is simply passed into the service being called.
 
+## Context Variables
+
+By Default, fetchr appends all context values to the xhr url as query params. `contextPicker` allows you to greater control over which context variables get sent as query params depending on the xhr method (`GET` or `POST`). This is useful when you want to limit the number of variables in a `GET` url in order not to accidentally [cache bust](http://webassets.readthedocs.org/en/latest/expiring.html).
+
+`contextPicker` follows the same format as the `predicate` parameter in [`lodash/object/pick`](https://lodash.com/docs#pick) with three arguments: `(value, key, object)`.
+
+var fetcher = new Fetcher({
+    context: { // These context values are persisted with XHR calls as query params
+        _csrf: 'Ax89D94j',
+        device: 'desktop'
+    },
+    contextPicker: {
+        GET: function (value, key, object) {
+            // for example, if you don't enable CSRF protection for GET, you are able to ignore it with the url
+            if (key === '_csrf') {
+                return false;
+            }
+            rerurn true;
+        }
+        // for other method e.g., POST, if you don't define the picker, it will pick the entire context object
+    }
+});
+
+var fetcher = new Fetcher({
+    context: { // These context values are persisted with XHR calls as query params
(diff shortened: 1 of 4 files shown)


mridgway:
:+1: Looks good.

README.md
```diff
@@ -302,6 +302,37 @@ fetcher
 
 For requests from the server, the config object is simply passed into the service being called.
 
+## Using Context Picker
+
+By Default, fetchr appends all context values to the xhr url, `contextPicker` allows you to determine how do you want to pick the context value. They should be the `predicate` function of [lodash/object/pick](https://lodash.com/docs#pick) with three arguments: `(value, key, object)`.
```
redonkulus: predicate "argument"

README.md
```diff
@@ -302,6 +302,37 @@ fetcher
...
+    }
+});
+
+var fetcher = new Fetcher({
+    context: { // These context values are persisted with XHR calls as query params
+        _csrf: 'Ax89D94j',
+        device: 'desktop'
+    },
+    contextPicker: {
+        GET: ['device'] // predicate can be an array of strings
```
redonkulus: this means just use `device`?

kaesonho: yeah it is

redonkulus:
:+1:

README.md
```diff
@@ -302,6 +302,37 @@ fetcher
 
 For requests from the server, the config object is simply passed into the service being called.
 
+## Using Context Picker
```
Vijar: Can we change this to "Context Variables"

README.md
Vijar: By Default, fetchr appends all context values to the xhr url as query params. `contextPicker` allows you to greater control over which context variables get sent as query params depending on the xhr method (`GET` or `POST`). This could be useful when you want to limit the number of variables in a `GET` url in order not to accidentally [cache bust](http://webassets.readthedocs.org/en/latest/expiring.html).

`contextPicker` follows the same format as the `predicate` parameter in [`lodash/object/pick`](https://lodash.com/docs#pick) with three arguments: `(value, key, object)`.

README.md
Vijar: We don't talk about context variables in depth anywhere else so this would be a good section for that.

kaesonho: Sure so this section becomes how do we use context in url and how do we control that using contextPicket.

kaesonho: updated as what you suggest above :+1:

Vijar:
:ship:

PR #122, opened 2015-09-18
Refactor service metadata

Vijar:
@redonkulus @mridgway 

Move logic from `fluxible-plugin-fetchr` into here and refactor. https://github.com/yahoo/fluxible-plugin-fetchr/pull/48

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -56,6 +56,9 @@ function pickContext (context, picker, method) {
  * @param {String} operation The CRUD operation name: 'create|read|update|delete'.
  * @param {String} resource name of fetcher/service
  * @param {Object} options configuration options for Request
+ * @param {Array} [options._serviceMeta] Array to hold per-request/session metadata from all service calls.
+ * Data will be pushed on to this array while the Fetchr instance maintains the reference for this session.
+ *
  * @constructor
  */
 function Request (operation, resource, options) {
@@ -70,7 +73,8 @@ function Request (operation, resource, options) {
         xhrTimeout: options.xhrTimeout || DEFAULT_XHR_TIMEOUT,
         corsPath: options.corsPath,
         context: options.context || {},
-        contextPicker: options.contextPicker || {}
+        contextPicker: options.contextPicker || {},
+        _serviceMeta: options._serviceMeta || []
     };
     this._params = {};
     this._body = null;
@@ -128,9 +132,16 @@ Request.prototype.end = function (callback) {
         setImmediate(executeRequest, self, resolve, reject);
     });
 
+    promise.then(function (result) {
+        if (result.meta) {
+            self.options._serviceMeta.push(result.meta)
+        };
+        return result;
+    });
+
     if (callback) {
-        promise.then(function (data) {
-            setImmediate(callback, null, data);
+        promise.then(function (result) {
+            setImmediate(callback, null, result.data, result.meta);
         }, function (err) {
             setImmediate(callback, err);
         });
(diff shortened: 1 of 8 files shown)


Vijar:
Btw, I haven't published promises yet either since we can't really support promises with `fluxible-plugin-fetchr` until after this gets merged. So I'll publish fetchr after.

libs/fetcher.client.js
```diff
@@ -128,9 +130,17 @@ Request.prototype.end = function (callback) {
         setImmediate(executeRequest, self, resolve, reject);
     });
 
+    promise.then(function (result) {
+        console.log('META', result);
```
kaesonho: you might want to remove this

Vijar: :) thanks

libs/fetcher.js
```diff
@@ -303,7 +318,10 @@ Fetcher.middleware = function () {
                         res.status(errResponse.statusCode).json(errResponse.output);
                         return;
                     }
-                    res.status(meta.statusCode || 200).json(data);
+                    res.status(meta.statusCode || 200).json({
```
redonkulus: does this break BC?

Vijar: Not really.

This PR also refactors the client fetchr so it parses out the `data` and `meta` separately. So as long as the middleware is only accessed by fetchr client calls and not directly via urls( which we told people to stop doing in the [last minor](https://github.com/yahoo/fetchr/releases/tag/v0.5.0)) this isn't a breaking change.

libs/fetcher.client.js
```diff
@@ -255,6 +264,7 @@ Request.prototype._constructGroupUri = function (uri) {
 
 function Fetcher (options) {
     this.options = options || {};
+    this.options.serviceMeta = [];
```
redonkulus: add this to FetchrClient doc block too

Vijar: Users don't pass this in, its just stored here.

redonkulus: Then probably should be this._serviceMeta so it's clear.

libs/fetcher.js
Vijar: So this could break BC if the client hasn't refreshed the page but the server gets updated to latest.

So I'm going to put in a feature flag to control this behavior and we can remove it in future versions.

libs/fetcher.client.js
Vijar: Done

Vijar:
@redonkulus @mridgway @kaesonho updated to be backwards compatible.

redonkulus:
LGTM. Lets make sure we test this with an app that has the old client code but new code on the server, just to ensure there are no BC issues.

Vijar:
I manually tested this with old client/new server combo and it works like a charm :)

Vijar:
@mridgway can you take one look at it.

libs/fetcher.client.js
```diff
@@ -171,6 +182,18 @@ function executeRequest (request, resolve, reject) {
             // TODO: Add test for this fallback
             get_uri = defaultConstructGetUri.call(request, uri, request.resource, request._params, clientConfig, request.options.context);
         }
+        // TODO: Remove `returnMeta` feature flag after next release
```
Vijar: This is the feature flag to prevent BC between versions.

libs/fetcher.client.js
```diff
@@ -171,6 +182,18 @@ function executeRequest (request, resolve, reject) {
...
+        // Whereas before any data from services was returned as is. We now return
+        // an object with a data key containing the service response, and a meta key
+        // containing the service's metadata response (i.e headers and statusCode).
+        // We need this feature flag to be truly backwards compatible because it is
+        // concievable that some active browser sessions could have the old version of
+        // client fetcher while the server upgrades to the new version. This could be
+        // easily fixed by refreshing the browser, but the feature flag will ensure
+        // old fetcher clients will receive the old format and the new client will
+        // receive the new format
+        get_uri = get_uri + '&returnMeta=true';
```
mridgway: Is it guaranteed that `?` and one other query param will always be present? Might be safer to add this to `request._params` before calling `defaultConstructGetUri`.

Vijar: @mridgway good point, I moved it from the query params into matrix params.

The only potential problem is if users' custom constructGetUri function somehow skips over the `returnMeta` key. This is a very edge condition though.

mridgway:
I didn't realize `params` was matrix. I feel like it does belong on query. Maybe it should just always be part of the `xhrContext` by default.

Vijar:
I actually put it in xhrContext but realized an edge case where it might break when people use customGetUri functions which might ignore serializing the context. 

This is why I was manually trying to add the query param to the final url, but like you pointed out it isn't easy to determine if it should be `?returnMeta=true` or `&returnMeta=true`. Perhaps the only sure way to do this is to deconstruct and then reconstruct the url with returnMeta added properly?

Vijar:
@mridgway updated.

mridgway:
:+1:

PR #125, opened 2015-10-01
chain promise to avoid bluebird's "Unhandled rejection" error

lingyan:
@Vijar @mridgway 

bluebird prints "Unhandled rejection" error when the promise is rejected without a rejected handler: https://github.com/petkaantonov/bluebird/blob/master/src/captured_trace.js#L267

fetchr.js currently uses `global.Promise`, which could be set by app to bluebird.  So to suppress this bluebird error, we need to register a no-op rejected handler.

Alternative is to simply use es6-promise in fetchr, instead of global.Promise and fallback to es6-promise.

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -132,7 +132,7 @@ Request.prototype.end = function (callback) {
         setImmediate(executeRequest, self, resolve, reject);
     });
 
-    promise.then(function (result) {
+    promise = promise.then(function (result) {
         if (result.meta) {
             self.options._serviceMeta.push(result.meta)
         };
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -137,7 +137,7 @@ Request.prototype.end = function (callback) {
         setImmediate(executeRequest, self, resolve, reject);
     });
 
-    promise.then(function (result) {
+    promise = promise.then(function (result) {
         if (result.meta) {
             self.serviceMeta.push(result.meta)
         };


mridgway:
In which case wouldn't there be a handler? If there is a callback, an error handler is added here: https://github.com/yahoo/fetchr/pull/125/files#diff-280ffc252427e4144d5ad8405f1c9e9dR155 and if the user is using promises, they should be registering their own handler. This seems like it's working as intended to me.

lingyan:
Yup.  That was what confused me at first as well.  It is an interesting implementation of bluebird.  When you have:
```js
var p = new Promise(function () {});
p.then(resolvedHandler);
p.then(resolvedHandler, rejectedHandler);
```
is different from real chaining like this:
```js
var p = new Promise(function () {});
p.then(resolvedHandler).then(resolvedHandler, rejectedHandler);
```

You can try running this test:
```js
global.Promise = require('bluebird');
var p = new global.Promise(function (resolve, reject) {
   reject('error--xyz');
});
p.then(function (result) {
   console.log('result=', result);
});
p.then(function (result) {
   console.log('result1=', result);
}, function (err) {
   console.log('err1=', err);
});
```
 You'll get this error:
```
$ node promise.js 
err1= error--xyz
Unhandled rejection Error: error--xyz
    at Object.ensureErrorObject (/Users/lingyan/node_modules/bluebird/js/main/util.js:261:20)
    at Promise._rejectCallback (/Users/lingyan/node_modules/bluebird/js/main/promise.js:469:22)
    at /Users/lingyan/node_modules/bluebird/js/main/promise.js:486:17
    at /Users/lingyan/tests/nodejs/promise.js:4:4
    at tryCatcher (/Users/lingyan/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._resolveFromResolver (/Users/lingyan/node_modules/bluebird/js/main/promise.js:480:31)
    at new Promise (/Users/lingyan/node_modules/bluebird/js/main/promise.js:70:37)
    at Object.<anonymous> (/Users/lingyan/tests/nodejs/promise.js:3:9)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:901:3
```

Vijar:
Can we just do 'promise = promise.then(//meta)` so we have proper chaining? That's what I meant to do originally.

lingyan:
@Vijar We could do that.  I wasn't sure if you wanted to do the original way for a reason.

lingyan:
@Vijar updated

kaesonho:
:+1:  @roderickhsiao  this might be the issue you mentioned before.

roderickhsiao:
:+1: nice

Vijar:
@lingyan can you also fix this in the client side promises: https://github.com/yahoo/fetchr/blob/3f950dfc7079fe648893580b6cc03af29ae36630/libs/fetcher.client.js#L135

lingyan:
@Vijar done

Vijar:
:ship:

mridgway:
:+1:

PR #126, opened 2015-10-15
Return promises in functions performing create, read, update, delete operations when there is no callback

webcarrot:


Vijar:
@webcarrot can you explain your reasoning for this PR?

Fetchr already returns promises in the CRUD methods when you use the proper syntax.

mridgway:
Seems reasonable to allow Promises for the old API as well. I don't think we'll be removing it any time soon.

webcarrot:
@Vijar returning Promise when there is no callback is common behavior

PS. Sorry for my English

Vijar:
Sorry @webcarrot I guess I didn't understand the usecase. Looks like this PR enables `fetchr.read('asdf', {id: 1}).then(...)` instead of `fetchr.read('asdf').params({id:1}).then(...)`.

I'm okay with merging this.

Vijar:
Although, writing tests for this would be a pain. @mridgway do you think we need to add tests for this or just silently support this feature and don't even document it?

PR #130, opened 2015-11-05
Allow setting of headers for erroneous responses

banghui:
Up till version 0.5.17 erroneous responses were able to carry meta-data. Is it possible to reinstate this capability in new versions of fetchr? Or is there a reason for disallowing meta-data from being passed on when the service fails? Thanks!

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -59,7 +59,8 @@ function getErrorResponse(err) {
 
     return {
         statusCode: statusCode,
-        output: output
+        output: output,
+        meta: err.meta
     };
 }
 
@@ -148,6 +149,9 @@ Request.prototype.end = function (callback) {
         promise.then(function (result) {
             setImmediate(callback, null, result.data, result.meta);
         }, function (err) {
+            if (err.meta) {
+                self.serviceMeta.push(err.meta);
+            };
             setImmediate(callback, err);
         });
     } else {
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -140,12 +140,12 @@ Request.prototype.end = function (callback) {
     promise = promise.then(function (result) {
         if (result.meta) {
             self.serviceMeta.push(result.meta);
-        };
+        }
         return result;
     }, function(errData) {
         if (errData.meta) {
             self.serviceMeta.push(errData.meta);
-        };
+        }
         throw errData.err;
     });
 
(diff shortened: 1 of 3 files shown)


redonkulus:
This looks good to me. Want to add a test?

Vijar:
@banghui I'm not able to find when we supported meta for error objects. I went to the code for `0.5.17` and am still not able to see it.

Perhaps adding a test case will clarify this for me.

redonkulus:
@banghui ping

banghui:
Sorry I forgot about this, thanks for the ping @redonkulus. 
@Vijar You are right. In `0.5.17` meta was never supported. We were, however, able to set  __headers__ for error responses then, but not anymore since the `0.5.18` release. I am trying to enable the setting of headers for error responses in this PR (not support meta in the error response payload, my bad!!). Thanks for clarifying!

In `0.5.17` headers were being set using the `meta` object provided by the [callbacks](https://github.com/yahoo/fetchr/blob/v0.5.17/libs/fetcher.js#L261-L265). In `0.5.17` headers were set using the [Request object's](https://github.com/yahoo/fetchr/blob/v0.5.18/libs/fetcher.js#L304-L314) `serviceMeta` property. The shift to using promises in `0.5.18` [omitted the setting of the Request object's serviceMeta property on errors](https://github.com/yahoo/fetchr/blob/v0.5.18/libs/fetcher.js#L141-L156), hence the inability to set headers for erroneous responses.

Let me know how I should proceed, thanks

Vijar:
@banghui thanks for clarifying. This is indeed an oversight on our part.

I think the fix would be to [return meta in errors here](https://github.com/yahoo/fetchr/blob/master/libs/fetcher.js#L168):

```js
        if (err) {
            reject({
                err: err,
                meta: meta
            });
        } else {
            resolve({
                data: data,
                meta: meta
            });
        }
```

Then change this [.then() function](https://github.com/yahoo/fetchr/blob/master/libs/fetcher.js#L140-L145) to:

```js
    promise = promise.then(function (result) {
        if (result.meta) {
            self.serviceMeta.push(result.meta)
        };
        return result;
    }, function (errData) {
        if (errData.meta) {
            self.serviceMeta.push(errData.meta);
        }
        throw errData.err;
    };
```

banghui:
@Vijar. Changes made. is this ok?

Vijar:
@banghui changes look good, do you think you can add a test for this? The coverage just dropped.

redonkulus:
ping @banghui

banghui:
added tests. sorry for the long delay :pray:

redonkulus:
LGTM. Will let @Vijar review before merging.

Vijar:
:+1:

PR #132, opened 2015-12-01
avoid sending undefined params in the uri

### LINKED ISSUE #55: coerce 'undefined' string to undefined
When a param is undefined, server side the api returns the actual undefined primitive, but on request from client the undefined param is now a string. This means for every string param we need to check whether the value is true-y and as well not equal to the string 'undefined'.

AaronLasseigne:
fixes #55 

Params that are `undefined` are being passed in the uri and then returned as the string `"undefined"`. It seems more appropriate to avoid sending any params that are `undefined`.

### DIFF
--- a/libs/util/defaultConstructGetUri.js
+++ b/libs/util/defaultConstructGetUri.js
@@ -39,7 +39,7 @@ module.exports = function defaultConstructGetUri(baseUri, resource, params, conf
         lodash.forEach(params, function eachParam(v, k) {
             if (k === id_param) {
                 id_val = encodeURIComponent(v);
-            } else {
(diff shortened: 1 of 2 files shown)


Vijar:
@AaronLasseigne Thanks for your PR! I really like it.

@redonkulus do you think this can be considered a breaking change or should we just publish?

redonkulus:
Seems like it should be fine, I don't know why apps would be relying on query params that are undefined.

Vijar:
Yeah, this is more of a bugfix anyway.

AaronLasseigne:
:+1:

PR #133, opened 2015-12-03
Add responseFormatter support

redonkulus:
@mridgway @Vijar @lingyan 

This allows applications the ability to add additional information to the XHR response before it is sent to the client. A typical use case might be to add debug information or performance metrics for XHR requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
-# Fetchr 
+# Fetchr
 
 [![npm version](https://badge.fury.io/js/fetchr.svg)](http://badge.fury.io/js/fetchr)
 [![Build Status](https://travis-ci.org/yahoo/fetchr.svg?branch=master)](https://travis-ci.org/yahoo/fetchr)
 [![Dependency Status](https://david-dm.org/yahoo/fetchr.svg)](https://david-dm.org/yahoo/fetchr)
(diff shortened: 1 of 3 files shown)


Vijar:
:+1:

README.md
```diff
@@ -258,6 +258,29 @@ fetcher
     });
 ```
 
+## XHR Response Formatting
+
+For some applications, there may be a situation where you need to modify an XHR response before it is passed to the client. Typically, you would apply your modifications in the service itself. However, if you want to modify the XHR responses across many services (i.e. add debug information), then you can use the `responseFormatter` option.
+
+`responseFormatter` is a function that is passed into the `Fetcher.middleware` method. It is passed three arguments, the request object, response object and the service response object from your service (i.e. the data returned from your service). The `responseFormatter` function can then modify the service response to add additional information.
```
lingyan: minor: and the <strike>service</strike> response object

libs/fetcher.js
```diff
@@ -279,12 +279,20 @@ Fetcher.isRegistered = function (name) {
  * Returns express/connect middleware for Fetcher
  * @method middleware
  * @memberof Fetcher
+ * @param {Object} [options] Optional configurations
+ * @param {Function} [options.responseFormatter=no op function] Function to modify the response
+            before sending to client. First argument is the request object,
+            second argument is the data object.
```
lingyan: I think there are 3 arguments.

lingyan:
minor comments on docs.  code is good.

PR #134, opened 2016-01-02
Clarify type of output property in README

c089:
The phrasing "contains valid JSON" sounded to me like this should be a JSON string.


Vijar:
Thanks for clearing that up.

PR #135, opened 2016-01-13
Security related fixes

lingyan:
@Vijar @redonkulus 

* escape resource name in error output

Based on [this owasp guideline](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#RULE_.233.1_-_HTML_escape_JSON_values_in_an_HTML_context_and_read_the_data_with_JSON.parse), 
```
Ensure returned Content-Type header is application/json and not text/html.
This shall instruct the browser not misunderstand the context and execute injected script.
```

Escaping is not needed for JSON data output, as long as the content-type header is set correctly to `Content-Type: application/json; charset=utf-8`, which fetchr already does with `res.json()`.

But there could be legacy browsers with security bugs that do not respect this, such as Opera browser prior to Opera 11.65  (Latest Opera is at 34.0 as of now.): https://www.cvedetails.com/cve/CVE-2012-3557/  

This PR escapes only the message strings fetchr library generates that can be part of the JSON response for addressing legacy browser concerns.  If application is concerned about legacy browser in general, the suggestion is to escape output app generates in the app before sending to fetchr.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,6 @@
+*.yml
+.editorconfig
+/docs/
 /artifacts/
 /examples/
 /tests/
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -12,6 +12,7 @@ var debug = require('debug')('Fetchr');
 var fumble = require('fumble');
 var objectAssign = require('object-assign');
 var Promise = global.Promise || require('es6-promise').Promise;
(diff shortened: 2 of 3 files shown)


redonkulus:
:+1:

Vijar:
:+1:

Issue #137, opened 2016-01-22
lodash 4 upgrade discussion


localnerve:
I just did an upgrade to lodash 4, however there is one incompatibility that bears discussion.
In lodash 4, pick is split into pick/pickBy where pickBy takes a predicate. However, the new predictate only receives a value*, where it used to receive (value, key, object).

This breaks contextPicker as-is.

##### Some Ideas
1. Pass the change* along
Just break the API, conform to the new lodash API, and change the contextPicker signature to just accepting a single value. Let the user decide accordingly.

2. Preserve the current API
There are a lot of ways to do this, and they're all kind of yucky (but work). They involve either wrapping the given options.contextPicker in a Fetchr supplied predicate that is given to lodash, or substituting a different method (for pick) entirely.

*For the key omission, looks like [this discussion](https://github.com/lodash/lodash/issues/1765) ended with [this update](https://github.com/lodash/lodash/commit/3eaa7b3b18069bfd32b98096c337edbb2cdca5fe). This should return (value, key) to the contextPicker.
Supplying the object is more trivial.

I anticipate a publish of 4.0.1 soon, so since this isn't urgent, I vote to:
1. Wait for [this update](https://github.com/lodash/lodash/commit/3eaa7b3b18069bfd32b98096c337edbb2cdca5fe), pickBy(value, key), to show up in the published package (presumably 4.0.1).
2. Pass the change along as-is to the API (the contextPicker signature update).

Thoughts?

PR #138, opened 2016-01-27
lodash 4.0.1 upgrade

localnerve:
Trying to help out. Upgraded to lodash 4.0.1, small adjustment made to contextPicker, added tests, updated readme. For history see issue #137 . The TLDR; is contextPicker no longer accepts an object as the third arg.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fetchr",
-  "version": "0.6.0",
+  "version": "0.5.25",
   "description": "Fetchr augments Flux applications by allowing Flux stores to be used on server and client to fetch data",
   "main": "index.js",
   "browser": "./libs/fetcher.client.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fetchr",
-  "version": "0.5.24",
+  "version": "0.6.0",
   "description": "Fetchr augments Flux applications by allowing Flux stores to be used on server and client to fetch data",
   "main": "index.js",
   "browser": "./libs/fetcher.client.js",
@@ -24,7 +24,7 @@
     "debug": "^2.0.0",
     "es6-promise": "^3.0.2",
     "fumble": "^0.1.0",
-    "lodash": "^3.3.0",
+    "lodash": "^4.0.1",
     "object-assign": "^4.0.1",
     "xhr": "^2.0.0"
   },
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ For requests from the server, the config object is simply passed into the servic
 
 By Default, fetchr appends all context values to the xhr url as query params. `contextPicker` allows you to greater control over which context variables get sent as query params depending on the xhr method (`GET` or `POST`). This is useful when you want to limit the number of variables in a `GET` url in order not to accidentally [cache bust](http://webassets.readthedocs.org/en/latest/expiring.html).
 
-`contextPicker` follows the same format as the `predicate` parameter in [`lodash/object/pick`](https://lodash.com/docs#pick) with three arguments: `(value, key, object)`.
+`contextPicker` follows the same format as the `predicate` parameter in [`lodash/pickBy`](https://lodash.com/docs#pickBy) with two arguments: `(value, key)`.
 
 ```js
 var fetcher = new Fetcher({
@@ -384,12 +384,12 @@ var fetcher = new Fetcher({
         device: 'desktop'
     },
     contextPicker: {
-        GET: function (value, key, object) {
+        GET: function (value, key) {
             // for example, if you don't enable CSRF protection for GET, you are able to ignore it with the url
             if (key === '_csrf') {
                 return false;
             }
-            rerurn true;
+            return true;
         }
         // for other method e.g., POST, if you don't define the picker, it will pick the entire context object
     }
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -12,11 +12,12 @@
 var REST = require('./util/http.client');
 var debug = require('debug')('FetchrClient');
 var lodash = {
-        isFunction: require('lodash/lang/isFunction'),
-        forEach: require('lodash/collection/forEach'),
-        merge: require('lodash/object/merge'),
-        noop: require('lodash/utility/noop'),
-        pick: require('lodash/object/pick')
+        isFunction: require('lodash/isFunction'),
+        forEach: require('lodash/forEach'),
+        merge: require('lodash/merge'),
+        noop: require('lodash/noop'),
+        pickBy: require('lodash/pickBy'),
+        pick: require('lodash/pick')
     };
 var DEFAULT_GUID = 'g0';
 var DEFAULT_XHR_PATH = '/api';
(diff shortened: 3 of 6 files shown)


mridgway:
Thanks for taking the time to do this. I think it's ok to just pass that change to `pick` on to our API. I don't see too much of a reason that people would need access to that object since it's just a plain object.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fetchr",
-  "version": "0.5.24",
+  "version": "0.6.0",
```
redonkulus: Can you revert this change? Don't think we will need to bump minor.

localnerve: Ok just go to 0.5.25?

Vijar: @localnerve yeah, bump patch version only.

localnerve:
I have some info that might be helpful to someone as the result of this change.
It involves the shared lodash dependency with fumble.
I ran into a chance occurrence where this caused the app to not be able to run correctly on **Heroku**.

fumble (a depenendency of fetchr) also uses lodash. Fumble's lodash dep is pegged at v3. So prior to this change, it was sharing the lodash dependency with this module when it also used v3. So now, that can no longer happen.

So if you are using a cached node_modules (like default on Heroku) you will run into a runtime issue because fumble will not be able to load (due specifically to the namespace differences introduced in lodash 4).
FWIW, on Heroku, the node_module cache is disrupted on install (default buildback heroku/nodejs) when the node runtime signature changes. For example, if the last node runtime was 4.2.6, if you peg it to 4.2.4 on the next install, that will disrupt the node_module cache and dependencies will be installed correctly.

Vijar:
@localnerve this would only be a problem if someone updates fetchr only? without updating its dependencies? Otherwise, as far as I'm aware `npm install` should recursively update dependencies.

localnerve:
If you do the update with npm install in any normal fashion, you're good.

My comment is only for PaaS or other environments that cache node_modules and layer in updates on top. The Heroku default build pack for node does this (heroku/nodejs) to be "efficient", but gets spectacularly bad results.

For Heroku, when you update the engine version to other than the last one used (can be any other version), you bust it:
```
-----> Restoring cache
       Skipping cache restore (new runtime signature)
```

Otherwise, it just brings in whatever node_modules it built last time, layering any changes on top. In the Heroku case, it did NOT detect the top-level fluxible-plugin-fetchr change, but DOES detect the top-level lodash change. I'm not saying this is good, it is just what it does.

To demonstrate, Here's part of the Heroku deploy output from the "successful" [flux-react-example-sw](https://github.com/localnerve/flux-react-example-sw) Travis build and deploy that caused a runtime crash when I first made the change to lodash 4. Notice that it only updates items in the node_modules cache that it thinks might have changed (Does NOT detect fluxible-plugin-fetchr, but DOES detect top-level, shared lodash 4.2.0):

```
-----> Restoring cache
       Loading 2 from cacheDirectories (default):
       - node_modules
       - bower_components (not cached - skipping)
-----> Building dependencies
       Pruning any extraneous modules
       Installing node modules (package.json)
       react-spinner@0.4.0 node_modules/react-spinner
       
       sitemap-xml@0.1.0 node_modules/sitemap-xml
       
       fg-loadcss@0.2.4 node_modules/fg-loadcss
       
       react-swipe@3.1.1 node_modules/react-swipe
       └── swipe-js-iso-ln@2.2.0
       
       fontfaceobserver@1.5.4 node_modules/fontfaceobserver
       ├── closure-dom@0.1.1
       └── promis@1.1.2
       
       express-error-handler@1.0.1 node_modules/express-error-handler
       ├── connect-domain@0.5.0
       ├── json-stringify-safe@5.0.1
       └── mout@0.7.1
       
       lodash@4.2.0 node_modules/lodash
-----> Caching build
       Clearing previous node cache
       Saving 2 cacheDirectories (default):
       - node_modules
       - bower_components (nothing to cache)
-----> Build succeeded!
       ├── amqplib@0.4.0
       ├── body-parser@1.14.2
       ├── classnames@2.2.3
       ├── compression@1.6.1
       ├── connect-modrewrite@0.8.5
       ├── cookie-parser@1.4.1
       ├── csurf@1.8.3
       ├── debug@2.2.0
       ├── diff@2.2.1
       ├── es6-promise@3.0.2
       ├── express@4.13.4
       ├── express-error-handler@1.0.1 (git://github.com/localnerve/express-error-handler.git#15eda3d4ac45bc467d5cadeba84fcda39e1cbaa2)
       ├── fg-loadcss@0.2.4
       ├── fluxible@1.0.3
       ├── fluxible-addons-react@0.2.3
       ├── fluxible-plugin-fetchr@0.3.9
       ├── fluxible-router@0.4.5
       ├── fontfaceobserver@1.5.4
       ├── inherits@2.0.1
       ├── lodash@4.2.0
       ├── morgan@1.6.1
       ├── nconf@0.8.2
       ├── node-jsx@0.13.3
       ├── nodemailer@2.0.0
       ├── object-assign@4.0.1
       ├── react@0.14.7
       ├── react-addons-css-transition-group@0.14.7
       ├── react-dom@0.14.7
       ├── react-modal@0.6.1
       ├── react-spinner@0.4.0 (git://github.com/localnerve/react-spinner.git#a9cbfb94c38cb894e4f15bca01976bc7635d474c)
       ├── react-swipe@3.1.1 (git://github.com/localnerve/react-swipe.git#3a5c3bce28695d367885c0a6cc525a552a78cc73)
       ├── remarkable@1.6.1
       ├── serialize-javascript@1.1.2
       ├── serve-favicon@2.3.0
       ├── sitemap-xml@0.1.0 (git://github.com/localnerve/sitemap-xml.git#8f375d0253733b982762261a864819b25e56ed0e)
       ├── superagent@1.7.2
       ├── sw-toolbox@3.1.1
       ├── treo@0.5.1
       └── web-push@0.0.11
```

In this case, since it did not see an update to fluxible-plugin-fetchr, it did not update it on top of the cached version. It left fumble with no lodash dependency of its own, relying on the previously satisfied dependency of lodash, which is found at the top level (now lodash 4.2.0 - oops!). 

I verified by `ls -l` the remote slug node_modules, and sure enough fumble did not have a v3 lodash of its own.

Should look like this:
```
$ heroku run ls -l ./node_modules/fluxible-plugin-fetchr/node_modules/fetchr/node_modules/fumble/node_modules --app flux-react-example
Running ls -l ./node_modules/fluxible-plugin-fetchr/node_modules/fetchr/node_modules/fumble/node_modules on flux-react-example... up, run.3204
total 8
drwx------  3 u44589 dyno 4096 Feb  2 15:33 http-status
drwx------ 14 u44589 dyno 4096 Feb  2 15:33 lodash
```
Above, that lodash is v3.x just for fumble (as required).

Here's the output from remote `ls -l` from the app crashing at runtime (different app, but same relevant deps):
```
$ heroku run ls -l ./node_modules/fluxible-plugin-fetchr/node_modules/fetchr/node_modules/fumble/node_modules --app pwa-stage
Running ls -l ./node_modules/fluxible-plugin-fetchr/node_modules/fetchr/node_modules/fumble/node_modules on pwa-stage... up, run.3130
total 4
drwx------ 3 u29762 dyno 4096 Jan 29 05:13 http-status
```
Notice no lodash at all. That is because it wasn't updated at all. Fumble's lodash dependency is still satisfied by the top level lodash, but unfortunately it is now 4.2.0, causing a runtime crash and violating fumble's requirements.

Hope this helps to save time responding to any "issues" and thank you again for this great, great work.

localnerve:
@Vijar forgot to `at` you initially, so you might not have received my previous reply in https://github.com/yahoo/fetchr/pull/138. This should help.

PR #140, opened 2016-03-31
use 500 as default status code for errors, instead of 400

lingyan:
400 is for client error.  Propose to use 500 instead.  Theoretically, this is a breaking change.  We can make it a non breaking change, by allowing people to configure default error code to return.  But don't think it is worth the trouble.  We can go 1.0 with this :)

@Vijar @mridgway @redonkulus

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ fetcher.updateOptions({
 
 ## Error Handling
 
-When an error occurs in your Fetchr CRUD method, you should return an error object to the callback. The error object should contain a `statusCode` (default 400) and `output` property that contains a JSON serializable object which will be sent to the client.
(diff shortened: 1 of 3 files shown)


mridgway:
Weird. :+1:

mridgway:
I think there are other breaking changes that we would want to get in to 1.0. Is this actually a breaking change?

lingyan:
Yeah. That is why I said **theoretically** it is a breaking change, if some app monitors the error status code from fetcher.  This is not urgent though.  We can wait for other breaking changes to publish 1.0.

Vijar:
:1:

yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

redonkulus:
👍

lingyan:
Based on discussions, we will release this in a patch release, as a bug fix.

Issue #141, opened 2016-04-08
Output in Error objects


gpbl:
To handle errors I'm adding `output` to the Error object, and then pass it to the service's callback:

```js
create(req, resource, params, config, callback) {
    const err = new Error('bad');
    err.output = 'Hi there, I\'m an error!';
    callback(err);
}
```

On the client I receive the output in the `body` property of the error. I was expecting an `output` property here:

```js
// call service client-side...
fetchr.read('myService').end(err => 
    console.log(err.output);  // undefined    
    console.log(err.body);  // Hi there, I'm an error!
);
```

On the server I receive instead the output in the `output` property:

```js
// calling the the same service server-side
fetchr.read('myService').end(err => 
    console.log(err.body);  // undefined
    console.log(err.output);  // Hi there, I'm an error!
);
```

Why this difference between `err.body` and `err.output`? Since I've to call the service both server and client side, this is how I do now:

```js
fetchr.read('myService').end(err => 
    console.log(err.body || err.output); // Hi there, I'm an error!
);
```

Not sure if i'm doing it right or I'm missing something. E.g. why has been chosen `output` and not `body` as property name?

redonkulus:
Thanks for the bug, @Vijar please take a look when you get a chance.

gpbl:
thanks!

PR #143, opened 2016-05-05
Refactor Fetchr Tests

Vijar:
A preview of whats to come.

Fixes https://github.com/yahoo/fetchr/issues/141


yahoocla:
CLA is valid!

Vijar:
@redonkulus @mridgway @lingyan this is ready for merge.

libs/fetcher.js
```diff
@@ -335,7 +335,14 @@ Fetcher.middleware = function (options) {
                     }
                     if (err) {
                         var errResponse = getErrorResponse(err);
-                        res.status(errResponse.statusCode).json(responseFormatter(req, res, errResponse.output));
+                        if (req.query && req.query.returnMeta) {
```
redonkulus: @Vijar can you explain this part? will this break responseFormatter since output is not passed in directly?

libs/util/http.client.js
```diff
@@ -193,6 +193,10 @@ function io(url, options) {
             err = new Error(errMessage);
             err.statusCode = status;
             err.body = errBody || body;
+            if (err.body) {
```
redonkulus: should we add a deprecate warning in the next release for `body`?

Vijar: I don't see why we can't keep both.

libs/fetcher.js
lingyan: i think we are fine.  We are already doing req.query.returnMeta logic branching on line 348 below

libs/util/http.client.js
redonkulus: Do we need to update docs to make sure only one is mentioned?

libs/fetcher.js
Vijar: Yeah, we were doing this for normal requests but not for error requests so its kind of a bug.

libs/util/http.client.js
Vijar: We don't even have docs on handling errors, but I just added some.

redonkulus:
👍

lingyan:
👍

PR #144, opened 2016-05-05
Expose xhr object for aborts

### LINKED ISSUE #142: Return the request object from fetchr call
Return the request object that has been created while making a fetchr call. Similar to the following

```javascript
var req = 	fetchr
			    .read('data_service')
			    .params({id: ###})
			    .end(function (err, data, meta) {
			    // handle err and/or data returned from data fetcher in this callback
			    });
```

Incase if we have to abort the request, it can be easily done just by doing
```javascript
req.abort();
```

Currently, there is no way of aborting the request that has been made using fetchr. A better use case will be when using fetchr for a component like auto-suggestion box, where-in we can abort the fetchr service calls that are already made (as the user types) and only fetch the data for the final query which user has typed-in.

Vijar:
@redonkulus @mridgway @praveenbnsm

We actually can expose the xhr object with little effort, but it just won't work with our promise interface though.

Fixes #142

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,32 @@ fetcher
     });
 ```
 
+## XHR Object
+
+The xhr object is returned by the `.end()` method as long as you're *not* chaining promises.
+This is useful if you want to abort a request before it is completed.
+
+```js
+var req = fetcher
+    .read('someData')
+    .params({id: ###})
+    .end(function (err, data, meta) {
+        // err.output will be { message: "Not found", more: "meta data" }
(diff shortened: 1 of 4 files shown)


yahoocla:
CLA is valid!

redonkulus:
We should probably make a note of this in our docs as a limitation.

Vijar:
@redonkulus added docs

redonkulus:
👍

PR #145, opened 2016-05-09
Returning rejected Promise when error occurred

koichik:
if the Promise's catch handler returns just a value, the Promise becomes fulfilled. So, it should return a rejected Promises.

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -152,7 +152,9 @@ Request.prototype.end = function (callback) {
             debug('Executing request %s.%s with params %o and body %o', self.resource, self.operation, self._params, self._body);
             setImmediate(executeRequest, self, resolve, reject);
         });
-        promise = promise.then(captureMeta, captureMeta);
+        promise = promise.then(captureMeta, function (err) {
+            throw captureMeta(err);
+        });
         return promise;
     }
 };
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

Vijar:
good catch!

libs/fetcher.client.js
```diff
@@ -152,7 +152,9 @@ Request.prototype.end = function (callback) {
             debug('Executing request %s.%s with params %o and body %o', self.resource, self.operation, self._params, self._body);
             setImmediate(executeRequest, self, resolve, reject);
         });
-        promise = promise.then(captureMeta, captureMeta);
+        promise = promise.then(captureMeta, function (err) {
+            return Promise.reject(captureMeta(err));
```
Vijar: In our server implementation (https://github.com/yahoo/fetchr/blob/master/libs/fetcher.js#L153) we just re-throw the error so I think to be consistent you can just do`throw captureMeta(err);` here.

koichik:
@Vijar Fixed.

kevhuang:
@Vijar Any estimate on when will this be released?

redonkulus:
@kevhuang this is now released in version 0.5.29.

kevhuang:
@redonkulus thank you!

PR #148, opened 2016-06-28
support fetchr stats collection and monitoring

lingyan:
This allows apps to provide their custom function to collect stats for fetchr requests.

@yahoo/team-fluxible

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -77,6 +77,9 @@ function getErrorResponse(err) {
  * @param {Object} options configuration options for Request
  * @param {Object} [options.req] The request object from express/connect.  It can contain per-request/context data.
  * @param {Array} [options.serviceMeta] Array to hold per-request/session metadata from all service calls.
+ * @param {Function} [options.statsCollector] The function will be invoked with 1 argument:
+ *      the stats object, which contains resource, operation, params (request params),
+ *      statusCode, err, and time (elapsed time)
  * @constructor
  */
 function Request (operation, resource, options) {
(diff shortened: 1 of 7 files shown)


yahoocla:
CLA is valid!

libs/fetcher.js
```diff
@@ -130,33 +135,59 @@ Request.prototype.clientConfig = function (config) {
...
+            resource: self.resource,
+            operation: self.operation,
+            params: self._params,
+            statusCode: err ? err.statusCode : (result && result.meta && result.meta.statusCode || 200),
+            err: err,
+            time: Date.now() - self._startTime
+        };
+        statsCollector(stats);
+    }
+};
```
lingyan: `err` and `result` objects do not have the same structure on client and server.  So had to duplicate this function here to keep code simple.  Looks like `Request` classes on server and client share a lot of common things.  We should refactor `fetchr.js` and `fetchr.client.js` to pull out common code and normalize data structure when we get time.

redonkulus: I thought the `err` and `result` objects were recently aligned with BC to ensure we didn't break anyone. But I agree the two files needs to be merged, its too easy to break now.

redonkulus:
Code looks good to me 👍

libs/fetcher.js
lingyan: We aligned `err` and `result` for the callback of `Request.prototype.end()`, not the `err` and `result` from `executeRequest` (private method for executing request).

PR #149, opened 2016-06-29
Made sure withCredentials config is passed along and not black-holed/…

activist:
…swallowed in mergeConfig.

Also needed to add withCredentials in the mergeConfig function which is kinda misleading name since it doesn't merge the two config as expected, but rather merges config specified with the function and overwrites the rest when returning.

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -117,6 +117,10 @@ function mergeConfig(config) {
             }
         }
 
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

lingyan:
👍

redonkulus:
👍

PR #150, opened 2016-07-20
be defensive. result could be null

lingyan:
`result` is parsed by [`parseResponse()`](https://github.com/yahoo/fetchr/compare/defensive?expand=1#diff-84c4f8c88b61ef0647afbcf02f0ae99cR30), which could return `null`


yahoocla:
CLA is valid!

kaesonho:
👍

PR #152, opened 2016-08-19
Check whether the number is valid in params in order to avoid Infinity

### LINKED ISSUE #151: `1234e1234` is parsed to Infinity
https://github.com/yahoo/fetchr/blob/master/libs/fetcher.js#L20

When the `1234e1234` was sent from client to server/middleware as param, it will be parsed to `Infinity`.

kouhin:
fix #151

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -18,12 +18,11 @@ function parseValue(value) {
     // take care of value of type: array, object
     try {
         var ret = JSON.parse(value);
-        // Numbers larger than MAX_SAFE_INTEGER will contain rounding errors so,
-        // we will just leave them as strings instead. The length > 15 check is because
-        // the MAX_SAFE_INTEGER in javascript is 9007199254740991 which has a length of 16.
-        // Using a length > 15 check is still safe and faster than trying to use another
-        // library that supports big integers.
-        if (typeof ret === 'number' && value.length > 15) {
+        // Big interger, big decimal and the number in exponential notations will results
+        // in unexpected form. e.g. 1234e1234 will be parsed into Infinity and the
+        // number > MAX_SAFE_INTEGER will cause a rounding error.
+        // So we will just leave them as strings instead.
+        if (typeof ret === 'number' && String(value) !== String(ret)) {
             ret = value;
         }
         return ret;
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

libs/fetcher.js
```diff
@@ -23,7 +23,7 @@ function parseValue(value) {
         // the MAX_SAFE_INTEGER in javascript is 9007199254740991 which has a length of 16.
         // Using a length > 15 check is still safe and faster than trying to use another
         // library that supports big integers.
-        if (typeof ret === 'number' && value.length > 15) {
+        if (typeof ret === 'number' && String(value) !== String(ret)) {
```
lingyan: can we just use `Number.isSafeInteger()`, which will also cover `Number.Infinity` case reported in #151? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger

kouhin: Yes, it seems that `isSafeInteger` is another solution. 
Up to now, I think `String(value)` is a simplest way to solve Infinity problem. Since it compares between original value and parsed value, it may also resolve other format problems, e.g. `JSON.parse('9007199254740991.11111')`

lingyan: Sure.  I'm ok with that.  Was also thinking another corner case like `123e1`, we probably don't really want to parse into number, at least at fetchr layer.

CC: @redonkulus @Vijar

redonkulus: @kouhin  I'm good with this. Please add a test.

libs/fetcher.js
```diff
@@ -23,7 +23,7 @@ function parseValue(value) {
         // the MAX_SAFE_INTEGER in javascript is 9007199254740991 which has a length of 16.
         // Using a length > 15 check is still safe and faster than trying to use another
         // library that supports big integers.
```
lingyan: @kouhin please update this doc block to reflect the code

lingyan:
@kouhin Please update the inline comment and add a unit test for this change.  Thanks!

kouhin:
I'm sorry that I'm very busy these days. 
I have updated the inline comment and add unit tests.
Thanks!

lingyan:
👍  Thanks @kouhin

lingyan:
Released in fetchr@0.5.35

PR #153, opened 2016-08-25
check service operation is defined

yosuke-furukawa:
`service[op].apply(service, args)` throws exception when `service[op]` is undefined.
And this task is called under `setImmediate`, so if service does not have the specified operation, `process.on('uncaughtException')` will be thrown. 

This PR is to avoid the above situation.

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -223,11 +223,13 @@ function executeRequest (request, resolve, reject) {
     var service;
     try {
         service = Fetcher.getService(request.resource);
+        if (!service[op]) {
+          throw new Error('operation: ' + op + ' is undefined on service: ' + request.resource);
+        }
+        service[op].apply(service, args);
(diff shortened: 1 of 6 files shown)


yahoocla:
CLA is valid!

lingyan:
👍

libs/fetcher.js
```diff
@@ -223,11 +223,13 @@ function executeRequest (request, resolve, reject) {
     var service;
     try {
         service = Fetcher.getService(request.resource);
+        if (!service[op]) {
+          throw new Error('operation: ' + op + ' is undefiend on service: ' + request.resource);
```
redonkulus: typo here: undefiend .. should be undefined

yosuke-furukawa:
Fixed.

redonkulus:
Thanks!

PR #154, opened 2016-08-30
prevent unsupported operation call

nysd:
I think fetcher.js should allow only 'read,create,update,delete' operations that fetchr.client supports.

because I could call any operation like 'constructor'. this may cause serious security problem.

```
curl -X POST http://localhost:3000/proxy -H "Content-Type:application/json" -d '{ "context":{},"requests":{ "g0": {"resource": "bookmarking", "operation": "constructor", "params":{}}}}' 
```

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -6,6 +6,7 @@
 var OP_READ = 'read';
 var OP_CREATE = 'create';
 var OP_UPDATE = 'update';
+var OP_DELETE = 'delete';
 var GET = 'GET';
 var qs = require('querystring');
 var debug = require('debug')('Fetchr');
@@ -223,7 +224,7 @@ function executeRequest (request, resolve, reject) {
     var service;
     try {
         service = Fetcher.getService(request.resource);
-        if (!service[op]) {
+        if ([OP_CREATE,OP_READ,OP_UPDATE,OP_DELETE].indexOf(op) < 0 || !service[op]) {
           throw new Error('operation: ' + op + ' is undefined on service: ' + request.resource);
         }
         service[op].apply(service, args);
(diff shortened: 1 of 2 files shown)


libs/fetcher.js
```diff
@@ -223,7 +224,7 @@ function executeRequest (request, resolve, reject) {
     var service;
     try {
         service = Fetcher.getService(request.resource);
-        if (!service[op]) {
+        if ([OP_CREATE,OP_READ,OP_UPDATE,OP_DELETE].indexOf(op) < 0 || !service[op]) {
```
lingyan: @nysd Thanks for your contribution!

How about moving this check before the line above, and checking using `===` explictly to avoid allocating an array and doing an indexOf every time request is executed?

yahoocla:
CLA is valid!

libs/fetcher.js
nysd: @lingyan Thanks for your advice.

I agree with you and I moved this check from 'executeRequest' to 'Request' function.

lingyan: @nysd Instead of moving this check to `Request` constructor and adding another try/catch, which is known to de-optimze the function for v8 engine, can we just do this check after line 431 the `if (!Fetcher.isRegistered(singleRequest.resource)) {` block, and call `next(error)` directly?  Thanks!

nysd: @lingyan Thanks for your advice. I moved the operation check from ```Request``` constructor to after ```!Fetchr.isRegistered``` block. Please confirm my code.

lingyan:
👍   Thanks! @nysd

lingyan:
Released in fetcher@0.5.34

PR #155, opened 2016-09-02
Add custom request header

nysd:
I would like to add custom request headers like 'X-VERSION'.
Usage is here.
```
const fetchr = new Fetchr({
    xhrPath: '/proxy',
    xhrTimeout: 3000,
    headers :{
      'X-VERSION': '1.0.0'
    }
  });
```

Because 'meta' like below  is hard to check at Nginx especially create/update/delete operation.
'meta' is not in request header but is in request body.
```
params : {
  meta : {
     header: {
        'X-VERSION': '1.0.0'
     }
  }
}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,8 @@
 sudo: false
 language: node_js
-matrix:
-  allow_failures:
-    - node_js: "0.13"
 node_js:
-  - "iojs"
-  - "0.13"
-  - "0.12"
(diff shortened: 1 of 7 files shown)


yahoocla:
CLA is valid!

lingyan:
@nysd  Code looks good.  Please add a section about customizing headers in README.  Then we are good to go.  Thanks!

PR #215, opened 2020-10-31
Add setImmediate polyfill

### LINKED ISSUE #195: setImmediate is not defined
Hi,
I often get setImmediate is not defined error, which is from [fetcher.client.js](https://github.com/yahoo/fetchr/blob/master/libs/fetcher.client.js).
I realized that [setImmediate](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) is not standard function that requires polyfill.
Can we replace it with setTimeout(fn, 0), which is the equivalent standard version, or [npm package](https://www.npmjs.com/package/setimmediate?activeTab=versions)?

Testing Environment
- Chrome 82
- Firefox 68.8
- Node.js 10 and 12

davisyi:
Add setImmediate polyfill﻿
Fixes #195

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -9,6 +9,7 @@
  * Fetcher is a CRUD interface for your data.
  * @module Fetcher
  */
+require("setimmediate");
 var REST = require('./util/http.client');
 var debug = require('debug')('FetchrClient');
 var lodash = {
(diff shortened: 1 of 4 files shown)


redonkulus:
thanks!

redonkulus:
Published in fetchr@0.5.40

PR #216, opened 2020-11-02
fetcher: rename service.name to service.resource

### LINKED ISSUE #51: Rename services' 'name' property to 'resource'
It makes more sense that way

pablopalacios:
Closes #51 

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,12 @@ var fetcher = new Fetcher({
 
 ### 3. Register data services
 
-You will need to register any data services that you wish to use in your application.
-The interface for your service will be an object that must define a `name` property and at least one [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) operation.
-The `name` propety will be used when you call one of the CRUD operations.
+You will need to register any data services that you wish to use in
+your application.  The interface for your service will be an object
+that must define a `resource` property and at least one
+[CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete)
+operation.  The `resource` property will be used when you call one of the
+CRUD operations.
 
 ```js
 // app.js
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -53,9 +53,12 @@ var fetcher = new Fetcher({
...
 ### 3. Register data services
 
-You will need to register any data services that you wish to use in your application.
-The interface for your service will be an object that must define a `name` property and at least one [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) operation.
-The `name` propety will be used when you call one of the CRUD operations.
+You will need to register any data services that you wish to use in
+your application.  The interface for your service will be an object
+that must define a `resource` property and at least one
+[CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete)
+operation.  The `resource` propety will be used when you call one of the
```
redonkulus, line 60:
```suggestion
operation.  The `resource` property will be used when you call one of the
```

libs/fetcher.js
```diff
@@ -283,11 +283,27 @@ Fetcher.registerFetcher = function (fetcher) {
...
     }
-    Fetcher.services[fetcher.name] = fetcher;
-    debug('fetcher ' + fetcher.name + ' added');
+
+    var resourceName;
+    if (typeof fetcher.resource !== 'undefined') {
+        resourceName = fetcher.resource;
+    } else if (typeof fetcher.name !== 'undefined') {
+        if ('production' !== process.env.NODE_ENV) {
+            console.warn('"fetcher.name" is deprecated. Please rename it to "fetcher.resource".');
```
redonkulus, line 297: This will be spammy in the console since we have many fetchers that use name. if we could only deprecate it once when the server starts, that would be better.

redonkulus (commented):
Looks good overall, just a couple of comments.

libs/fetcher.js
pablopalacios: I've moved this logic to `Fetcher.middleware`. I also added an array to control all the deprecated services definitions. In this way, it will be easier to spot services non compliant with the newer API.

redonkulus (approved):
LGTM, thanks!

Issue #251, opened 2021-06-16
Fetchr 0.5.42 and IE11


matthoffner:
Heads up we are noticing some test failures with IE11 and the recently released fetchr (0.5.42)

Error:

`Expected identifier, string or number`

Understandable that IE11 is on its way out but I wanted to surface this in case anyone else sees something similar. Ideally any change that drops IE11 support would be major, but we can pin for the time being.

redonkulus:
Yup, we hit this internally too, we ended up just updating our babel/webpack config to transpile fetchr as well. We should have changed this in a major version.

@pablopalacios if you want to refactor your latest change to remove ES6 that would help users that haven't discovered the issue yet.

pablopalacios:
Oh, it was really not my intention to drop support for IE11 since we also support it here. I tried to be very careful about it but it looks like some thing slipped in. I'll check it soon ;)

redonkulus:
@matthoffner this is now fixed via https://github.com/yahoo/fetchr/pull/252, released in fetchr@0.5.43.

Thanks for the report!

PR #252, opened 2021-06-17
fix ie11 regression on fetcher.client.js

pablopalacios:
The problem was on the following code:

```
    return { [p]: context[p] };
```

ie11 doesn't support this shortcut and sees it as syntax error. I wan the script on ie11 on win7 and I don't get any syntax error anymore.

Solves #251 

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -52,30 +52,25 @@ function pickContext (context, picker, method) {
     }
 
     var p = picker[method];
+    var result = {};
 
     if (typeof p === 'string') {
-        return { [p]: context[p] };
-    }
-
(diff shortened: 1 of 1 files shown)


libs/fetcher.client.js
```diff
@@ -52,30 +52,25 @@ function pickContext (context, picker, method) {
     }
 
     var p = picker[method];
+    var result = {};
 
     if (typeof p === 'string') {
-        return { [p]: context[p] };
```
pablopalacios, line 57: @redonkulus this was the problematic line.

redonkulus:
Thanks for fixing this quickly!

matthoffner:
Thanks @pablopalacios

PR #260, opened 2021-07-09
Improve retry mechanism

pablopalacios:
@redonkulus while thinking in a way of adding retry mechanism to fetchr to solve an issue in my company, I discovered that fetchr already has support for it. I went deeper on it and realized that there was a bug with the retry interval and that the status code check was only supporting 408 and xhr failures. Then I fixed the bug and add support for custom status while keeping it backward compatible.

- fix initial retry.interval not being respected
- add support for custom response status codes to be retried automatically

**UPDATE:**
- add full jitter formula for interval calculation
- deprecate `max_retries` in favor of `maxRetries` 
- add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,8 +432,7 @@ For requests from the server, the config object is simply passed into the servic
 
 ## Retry
 
-You can set Fetchr to retry failed requests automatically by setting a
-`retry` settings in the client configuration:
+You can set Fetchr to retry failed requests automatically by setting a `retry` settings in the client configuration:
 
 ```js
 fetcher
--- a/README.md
+++ b/README.md
@@ -446,9 +446,7 @@ fetcher
     .end();
 ```
 
-With this confiugration, Fetchr will retry all requests that fails
-with 408 status code or with a XHR 0 status code for 2 more times
-before returning an error. The interval between each requests respects
+With this configuration, Fetchr will retry all requests that fail with 408 status code or with an XHR 0 status code two more times before returning an error. The interval between each request respects
 the following formula, based on the exponential backoff and full jitter strategy published in [this AWS architecture blog post](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/):
 
 ```js
--- a/README.md
+++ b/README.md
@@ -430,6 +430,69 @@ fetcher.read('service').params({ id: 1 }).clientConfig(config).end(callbackFn);
 
 For requests from the server, the config object is simply passed into the service being called.
 
+## Retry
+
+You can set Fetchr to retry failed requests automatically by setting a
+`retry` settings in the client configuration:
+
+```js
+fetcher
+    .read('service')
+    .clientConfig({
+        retry: {
+            maxRetries: 2,
+        },
+    })
+    .end();
+```
+
+With this confiugration, Fetchr will retry all requests that fails
+with 408 status code or with a XHR 0 status code for 2 more times
+before returning an error. The interval between each requests respects
+the folowwing formula:
+
+```js
+Math.random() * Math.pow(2, attempt) * interval;
+```
+
+`attempt` is the number of the current retry attempt starting
+from 0. By default `interval` corresponds to 200ms.
+
+You can customize the retry behavior by adding more properties in the
+`retry` object:
+
+```js
+fetcher
+    .read('resource')
+    .clientConfig({
+        retry: {
+            maxRetries: 5,
+            interval: 1000,
+            statusCodes: [408, 502],
+        },
+    })
+    .end();
+```
+
+With the above configuration, Fetchr will retry for at maxiumum of 5
+times all the requests that fails with a 408 or 502 status code. The
+interval between each request will still use the formula from above,
+but the interval of 1000ms will be used instead.
+
+**Note:** Fetchr doesn't retry POST requests for safety reasons. You
+can enable retries for POST requests by setting the `unsafeAllowRetry`
+property to `true`:
+
+```js
+fetcher
+    .create('resource')
+    .clientConfig({
+        retry: { maxRetries: 2 },
+        unsafeAllowRetry: true,
+    })
+    .end();
+```
+
 ## Context Variables
 
 By Default, fetchr appends all context values to the xhr url as query params. `contextPicker` allows you to greater control over which context variables get sent as query params depending on the xhr method (`GET` or `POST`). This is useful when you want to limit the number of variables in a `GET` url in order not to accidentally [cache bust](http://webassets.readthedocs.org/en/latest/expiring.html).
--- a/README.md
+++ b/README.md
@@ -449,7 +449,7 @@ fetcher
 With this confiugration, Fetchr will retry all requests that fails
 with 408 status code or with a XHR 0 status code for 2 more times
 before returning an error. The interval between each requests respects
-the folowwing formula:
+the following formula, based on the exponential backoff and full jitter strategy published in [this AWS architecture blog post](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/):
 
 ```js
 Math.random() * Math.pow(2, attempt) * interval;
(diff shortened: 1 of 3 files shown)


libs/util/http.client.js
```diff
@@ -103,41 +107,34 @@ function mergeConfig(config) {
...
+            if (config.retry.statusCodes) {
+                cfg.retry.statusCodes = config.retry.statusCodes
+            }
         }
 
         if (config.withCredentials) {
             cfg.withCredentials = config.withCredentials;
         }
-
-        // tmp stores transient state data, such as retry count
```
redonkulus, line 120: Why was this removed?

libs/util/http.client.js
```diff
@@ -103,41 +107,34 @@ function mergeConfig(config) {
...
             if (!isNaN(interval) && interval > 0) {
                 cfg.retry.interval = interval;
             }
-            maxRetries = parseInt(config.retry && config.retry.max_retries, 10);
+            maxRetries = parseInt(config.retry.max_retries, 10);
             if (!isNaN(maxRetries) && maxRetries >= 0) {
                 cfg.retry.max_retries = maxRetries;
             }
+            if (config.retry.statusCodes) {
+                cfg.retry.statusCodes = config.retry.statusCodes
```
redonkulus, line 119: Lets document this new option in the readme please.

redonkulus (changes requested):
Thanks for the PR!

Looks like I need to run prettier on this code base, its showing its age. Left a couple of comments. Also tagged @lingyan to take a look since she might have more knowledge of this retry logic than me.

libs/util/http.client.js
pablopalacios: **Context**: for every retry, we double the interval for the next request. And, if the amount of retries reaches the maximum limit, `fetchr` just give up and returns an error.

**Previous implementation:** `cfg.tmp` was only being used to store the number of attempts and the interval for the next request. Whenever we would call `shouldRetry`, we would mutate this object by increasing the number of attempts and by doubling the interval so the next retry could be properly set.

 What I did was to replace this mutable object with an explicit required `attempt` param in `doXhr` function. `doXhr` is the function that is responsible for attempting the request, calling `shouldRetry` and scheduling a new request attempt if necessary (by calling itself). 

Since `doXhr` has `attempt` as param now, it can just pass `attempt + 1` to the inner `doXhr` call. Also, we can get the new interval value by just doing `Math.pow(2, attempt) * interval`. So no need for a mutable object that is shared between calls.

pablopalacios:
> Looks like I need to run prettier on this code base, its showing its age. 

That would be so awesome!

redonkulus:
Added here, this will mess up your PR once merged https://github.com/yahoo/fetchr/pull/261

libs/util/http.client.js
```diff
@@ -146,12 +143,14 @@ function doXhr(method, url, headers, data, config, callback) {
...
+                if (!shouldRetry(method, config, response.statusCode, attempt)) {
                     callback(err);
                 } else {
                     setTimeout(
-                        function retryXHR() { doXhr(method, url, headers, data, config, callback); },
-                        config.retry.interval
+                        function retryXHR() {
+                            doXhr(method, url, headers, data, config, attempt + 1, callback);
+                        },
+                        config.retry.interval * Math.pow(2, attempt)
```
lingyan, line 153: @pablopalacios since you are at this, do you mind improving this exponential retry logic to use full jitter, as documented [here](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) by aws?

lingyan (commented):
I like the changes! Logic is cleaner.  Left a comment for enhancement.  Thanks!

libs/util/http.client.js
pablopalacios: What should we use as cap? Or should we not use cap at all?

libs/util/http.client.js
pablopalacios: Done. But while documenting I felt that it was so wrong to use `max_retries` while all the other settings use camel case names that I added another commit deprecating `max_retries` in favor of `maxRetries`.

libs/util/http.client.js
pablopalacios: @lingyan I updated it but I haven't set `cap` for now so the delay will always be a random number between 0 and `                        config.retry.interval * Math.pow(2, attempt)`. My feeling was that setting a `cap` as in the aws article would require us to set another property in the configuration to allow users to tweak it properly if they have different interval configurations. We could instead, in the future, add another configuration where users can pass a custom function to calculate the delay.

lingyan: Sure!  I am ok with not using cap.  We can add it later if needed.  Thanks @pablopalacios for making the change!

libs/util/http.client.js
```diff
@@ -152,12 +155,27 @@ function doXhr(method, url, headers, data, config, callback) {
                 callback(NULL, response);
             },
             failure: function (err, response) {
-                if (!shouldRetry(method, config, response.statusCode)) {
+                if (
+                    !shouldRetry(method, config, response.statusCode, attempt)
+                ) {
                     callback(err);
                 } else {
+                    var delay =
```
lingyan, line 163:
```suggestion
                    // Use exponential backoff and full jitter strategy published in https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
                    var delay =
```

lingyan (commented):
Left two doc related suggestions. minor. rest looks good. Thanks!

lingyan (approved):

README.md
```diff
@@ -430,6 +430,69 @@ fetcher.read('service').params({ id: 1 }).clientConfig(config).end(callbackFn);
...
+            maxRetries: 2,
+        },
+    })
+    .end();
+```
+
+With this confiugration, Fetchr will retry all requests that fails
+with 408 status code or with a XHR 0 status code for 2 more times
+before returning an error. The interval between each requests respects
+the following formula, based on the exponential backoff and full jitter strategy published in [this AWS architecture blog post](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/):
```
redonkulus, line 452:
```suggestion
With this configuration, Fetchr will retry all requests that fail with 408 status code or with an XHR 0 status code two more times before returning an error. The interval between each request respects
the following formula, based on the exponential backoff and full jitter strategy published in [this AWS architecture blog post](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/):
```

README.md
```diff
@@ -430,6 +430,67 @@ fetcher.read('service').params({ id: 1 }).clientConfig(config).end(callbackFn);
...
+```
+
+With the above configuration, Fetchr will retry for at maxiumum of 5
+times all the requests that fails with a 408 or 502 status code. The
+interval between each request will still use the formula from above,
+but the interval of 1000ms will be used instead.
+
+**Note:** Fetchr doesn't retry POST requests for safety reasons. You
+can enable retries for POST requests by setting the `unsafeAllowRetry`
+property to `true`:
```
redonkulus, line 482:
```suggestion
With the above configuration, Fetchr will retry all failed (408 or 502 status code) requests for a maximum of 5 times. The interval between each request will still use the formula from above, but the interval of 1000ms will be used instead.

**Note:** Fetchr doesn't retry POST requests for safety reasons. You can enable retries for POST requests by setting the `unsafeAllowRetry` property to `true`:
```

README.md
```diff
@@ -430,6 +430,62 @@ fetcher.read('service').params({ id: 1 }).clientConfig(config).end(callbackFn);
 
 For requests from the server, the config object is simply passed into the service being called.
 
+## Retry
+
+You can set Fetchr to retry failed requests automatically by setting a
+`retry` settings in the client configuration:
```
redonkulus, line 436:
```suggestion
You can set Fetchr to retry failed requests automatically by setting a `retry` settings in the client configuration:
```

redonkulus (approved):

redonkulus:
Published https://github.com/yahoo/fetchr/releases/tag/v0.5.45

PR #261, opened 2021-07-09
chore: add prettier to project

redonkulus:


package.json
```diff
@@ -33,12 +33,15 @@
     "chai": "^4.2.0",
     "coveralls": "^3.0.5",
     "eslint": "^7.29.0",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-prettier": "^3.4.0",
```
pablopalacios, line 37: I would recommend you to not integrate prettier with eslint. Besides not being the recommended way according to prettier docs, you will not have to maintain 2 extra libraries that will couple eslint and prettier versions.

`eslint:recommended` preset doesn't stay in the way of prettier and running `prettier --check .` on CI is enough to prevent ugly code to reach master.

You could also scripts alias to format/check the code to make it easier to run prettier outside of an editor env:

```json
"scripts": {
    "format": "prettier --write .",
    "format:check": "prettier --check ."
}
```

pablopalacios:
noooo...

package.json
pablopalacios: too late :(

redonkulus: Oops haha, I can refactor it, I read that from the docs but stil felt easier to run it via eslint since we already had it setup.

redonkulus: https://github.com/yahoo/fetchr/pull/262

PR #263, opened 2021-07-11
http: remove trim polyfill

pablopalacios:
@redonkulus `String.prototype.trim` was added in IE9. IE8 was deprecated by MS [in 2016](https://www.microsoft.com/en-us/microsoft-365/windows/end-of-ie-support). IMO we could remove this polyfill without a need of a major release since people only need support for IE11 (but I might be wrong). What do you think?

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -30,13 +30,6 @@ var DEFAULT_CONFIG = {
 
 var forEach = require('./forEach');
 
-//trim polyfill, maybe pull from npm later
-if (!String.prototype.trim) {
-    String.prototype.trim = function () {
-        return this.replace(/^\s+|\s+$/g, '');
-    };
-}
-
 function normalizeHeaders(headers, method, isCors) {
     var normalized = {};
     if (!isCors) {


redonkulus:
We could probably modernize this whole library and remove < ie 11 support. I think forEach might be old too. I wouldn’t mind bumping major as well just to be safe.

redonkulus:
@pablopalacios obviously this is major scope creep but this would also be an opportunity to remove the deprecated methods from years ago and get this package to a 1.x release: https://github.com/yahoo/fetchr/issues?q=is%3Aissue+is%3Aopen+label%3A%22Breaking+Change%22

pablopalacios:
Exactly. But I'm afraid of another spiral of changes that never end as it happened with fluxible-addons-react. For me, right now, removing all the polyfills would be the most valuable thing so we can make fetchr very thin in the client. 

I've been refactoring the tests to not rely on mocking `xhr` so we can replace it with native `XMLHttpRequest`. As we discussed in another issue, we could also remove `es6-promise` polyfill. 

The only one remaining would be `setImmediate`. To be honest, I don't understand quite well its purpose and if it's still valid to use it nowadays since it didn't become a standard. I might be wrong but if I got it, the idea is to break long tasks in smaller ones to avoid blocking the main thread continuously. If so, perhaps we could just use more promises or, in the worse scenario, use `setTimeout` with 0 interval.

Removing all the polyfills would be quite cool since, even if we do a major breaking release, probably most of the people would be able to migrate to the newer version without any code change. The ones that still support old browsers would only have to install the polyfills by themselves (and probably they already have all them installed for other libraries or own code).

Said that, I think this would be already enough material for a 1.0 that could be finished/released quite soon.

---

Then comes the API itself. I support removing the callback interface and only sticking with promises. That would be awesome!

But regarding the new API, I'm not a big fan of it. I migrated some calls in my codebase and having to write `fetchr.read('resource').end().then(response => doSomething(response))` didn't feel nice. Ideally I would like to not need to call `end()`. Also, mocking these chained calls is quite annoying. So I opted in my code to keep using the old API so I don't have to rewrite the tests. 

In the other hand, the current signature is not so smart as well. If I just want to create something with a timeout I have to do `fetchr.create('resource', null, null, { timeout: 500 })`. It's easier to mock but super confusing when reading the implementation (besides the unnecessary typing).

My proposal would be to not have any chained method and just pass an options object with everything that is needed inside: `fethr.read('resource', { timeout: 300, })` It would be easy to mock but also would make clear which options are being passed (and wouldn't require unnecessary typing as the `end` call or the double `null`s). Therefore, I would leave this open for a v2 so you can also think how the API could be ideally.

redonkulus:
Lots to unpack here, will answer inline...

> Exactly. But I'm afraid of another spiral of changes that never end as it happened with fluxible-addons-react. For me, right now, removing all the polyfills would be the most valuable thing so we can make fetchr very thin in the client.

This is fine for a 0.6 release, I just wouldn't want to remove polyfills without releasing a minor version break (minor cause we are < 1.x).

> 
> I've been refactoring the tests to not rely on mocking `xhr` so we can replace it with native `XMLHttpRequest`. As we discussed in another issue, we could also remove `es6-promise` polyfill.
> 
> The only one remaining would be `setImmediate`. To be honest, I don't understand quite well its purpose and if it's still valid to use it nowadays since it didn't become a standard. I might be wrong but if I got it, the idea is to break long tasks in smaller ones to avoid blocking the main thread continuously. If so, perhaps we could just use more promises or, in the worse scenario, use `setTimeout` with 0 interval.

I'm fine removing `setImmediate`, when we used it, it was still in review, since its not standard we should refactor to something else (promises or setTimeout if need be). Against this should be part of 0.6 release since it could possibly break something.

> 
> Removing all the polyfills would be quite cool since, even if we do a major breaking release, probably most of the people would be able to migrate to the newer version without any code change. The ones that still support old browsers would only have to install the polyfills by themselves (and probably they already have all them installed for other libraries or own code).
> 
> Said that, I think this would be already enough material for a 1.0 that could be finished/released quite soon.

Yes, this could either be 0.6 or part of a 1.x release.

> 
> Then comes the API itself. I support removing the callback interface and only sticking with promises. That would be awesome!
> 
> But regarding the new API, I'm not a big fan of it. I migrated some calls in my codebase and having to write `fetchr.read('resource').end().then(response => doSomething(response))` didn't feel nice. Ideally I would like to not need to call `end()`. Also, mocking these chained calls is quite annoying. So I opted in my code to keep using the old API so I don't have to rewrite the tests.

You could just `await` instead of `.then()` but I get your point, calling `.end()` is odd in general. The `read` should just return the `Promise` directly. I can't remember the original purpose of `.end()`, perhaps to support the callback API.

> 
> In the other hand, the current signature is not so smart as well. If I just want to create something with a timeout I have to do `fetchr.create('resource', null, null, { timeout: 500 })`. It's easier to mock but super confusing when reading the implementation (besides the unnecessary typing).

Agree this interface is quite awkward.

> 
> My proposal would be to not have any chained method and just pass an options object with everything that is needed inside: `fethr.read('resource', { timeout: 300, })` It would be easy to mock but also would make clear which options are being passed (and wouldn't require unnecessary typing as the `end` call or the double `null`s). Therefore, I would leave this open for a v2 so you can also think how the API could be ideally.

Completely agree here, this would be part of a 1.x release since its a pretty dramatic change to the API.

The smaller polyfill removal changes could be a 0.6 release since its breaking but does not require users to do a full scale refactor.

pablopalacios:
Awesome! So I'll prepare the PRs removing the polyfills for the 0.6.

redonkulus:
@pablopalacios 0.6.0 has been published now with the 3 PRs you opened.

PR #273, opened 2021-07-29
http: remove unused put and delete methods

pablopalacios:
@redonkulus while refactoring the tests for the other PR, I realized that we don't use `delete` nor `put` from the http module.

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -27,9 +27,7 @@ var DEFAULT_CONFIG = {
     CONTENT_TYPE = 'Content-Type',
     TYPE_JSON = 'application/json',
     METHOD_GET = 'GET',
-    METHOD_PUT = 'PUT',
     METHOD_POST = 'POST',
-    METHOD_DELETE = 'DELETE',
     NULL = null;
 
 var INITIAL_ATTEMPT = 0;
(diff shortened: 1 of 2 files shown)


redonkulus:
@pablopalacios Can you explain this more? What do you mean they were not used?

pablopalacios:
Yes.

The flow in the client is something like this: 

1. Instantiate `Fetchr`
2. Call a operation method in a resource `fetchr.read('resource')`
3. Internally, fetchr it will create a Request object (https://github.com/yahoo/fetchr/blob/master/libs/fetcher.client.js#L450)
4. the request object has a method `end` that calls `executeRequest` (https://github.com/yahoo/fetchr/blob/master/libs/fetcher.client.js#L185)
5. `executeRequest` decides if it should use `POST` or `GET` based in the operation (https://github.com/yahoo/fetchr/blob/master/libs/fetcher.client.js#L185)
6. Based on that, we will either call http.client.get (https://github.com/yahoo/fetchr/blob/master/libs/fetcher.client.js#L295) or http.client.post (https://github.com/yahoo/fetchr/blob/master/libs/fetcher.client.js#L324)

You can check all the references for `http.client` (import as `REST`), only `get` and `post` methods are called in the whole codebase. So even though I would expect that `update` and `delete` methods would use `PUT` and `DELETE`, currently they will use `POST`.

If caching was not an important thing, I would make all the requests using `POST` and pass all the stuff in the request body.

redonkulus:
Good find, this is interesting. Do we not see the need for `put` and `delete` from the client code? I guess it's never come up before, so it's probably safe to remove.

pablopalacios:
I see no reason for using it. In the end the communication between client and server is a "closed" fetchr protocol: client fetchr talks to server fetchr middleware. As I said, we could even do everything using only POST for the sake of making things simpler. But this would harm caching or make it very difficult.

redonkulus (approved):

redonkulus:
https://github.com/yahoo/fetchr/releases/tag/v0.6.2

Issue #278, opened 2021-08-06
Abort support, retry mechanism and promises


pablopalacios:
@redonkulus 

**Context:** currently calling fetchr.read() with a callback [will return](https://github.com/yahoo/fetchr/blob/master/libs/util/http.client.js#L174) a `XMLHttpRequest` instance that allows us to `abort` the request if needed:

```js
const request = fetchr.read('resource', {}, {}, callback);
request.abort()
```

When we use the Promise API, this is not currently possible since we are returning a native promise. This is expected and documented since we don't implement a custom Promise interface to circumvent this issue.

**Issue:** While I was working on #271 I found another bug related to the the retry mechanism: `abort` doesn't work if the retry mechanism is triggered. The reason: [we are creating](https://github.com/yahoo/fetchr/blob/master/libs/util/http.client.js#L157) new `XMLHttpRequest` instances inside a callback and because of that the users don't have access to them. As a consequence, the request can't be aborted anymore since the one that was returned initially was already completed.

Then comes the dilemma: should we fix the issue by removing the abort feature or by really fixing it? If we kill it, it will make the job of converting everything to promises easier since we wouldn't have to create a custom promise interface that would allow us to abort it. In the other hand, `abort` seems to be a cool feature to have... We could also add it later if, for example, we replace xhr with `fetch` which would allow us to use the same `AbortController` for all requests that are triggered...

What do you say?

redonkulus:
I would prefer to migrate to native `fetch`, removing `xhr` and then using `AbortController`, since its a standard API.

pablopalacios:
#290 solved this issue.

Issue #275, opened 2021-07-29
Refactor fetchr API


redonkulus:
## Goal

Make the API more intuitive and easy to use for developers

## Background

Some of this work is based on the discussion in https://github.com/yahoo/fetchr/pull/263#issuecomment-881741395

### Simplify CRUD signatures

- Remove method chaining and just pass an options object with everything that is needed inside. It would be easy to mock and would make it clear which options are being passed (and wouldn't require unnecessary typing as the `end` call or the double `null`).
- Remove `callback` interface
- Remove `end()` method

Current:
```js
fetchr.create('resource', null, null, { timeout: 500 });
```

Proposal:
```js
fetchr.create('resource', { data: {}, timeout: 500 })
```

/cc @pablopalacios

lingyan:
I like the proposal.  A couple questions:

- Can you clarify what goes into `data` should be specified, I assume it is post data. 
- Original API's second argument (first `null`) is for resource params, and third argument (second `null`) is for request-specific configs.  Will they become fields in the second argument of the new API?

Would be good to add these details in the proposal.

redonkulus:
@lingyan

> I like the proposal. Some questions:
> 
> * Can you clarify what goes into `data`?  I assume it is post/put body.

Yup, that was intended to be post body data. We could figure out what the correct property name should be (`data`, `body`, `postBody`, etc). `body` is used in the code and documentation today.

> * Original API's second argument (first `null`) is for resource params, third argument (second `null`) is for post body, and last argument is for request-specific configs.  Will resource params become `params` in the second argument of the new API?

I think that makes sense, One potential option could be something like this (which basically removes the first param and puts it into the object):

```js
fetchr.create({ resource: 'foo', body: {}, params: {}, timeout: 500 });
```

NOTE: I don't like having `timeout` where it is, see comment below.

> * Are there example of other configs, besides `timeout`?
> 

We have them in the readme, the config that gets passed into the `clientConfig` method. Examples include `timeout`, `retry`, `unsafeAllowRetry`, `cors`, `constructGetUri`. So we may need to contain these in a `config` property or something more meaningful like `contextConfig` or `requestConfig`. Not sure on the name but something to make it clear its different from `params` and its per request.

> Would be good to add these details in the proposal.

pablopalacios:
I like the idea of moving everything to just one object. In the future, the `resource` property could also accept an array of resources so we could trigger many requests in the server. This is something that would be trickier to do with the current API.

[...]

PR #290, opened 2021-09-19
Replace xhr with native fetch

pablopalacios:
## What is done
- Completely replaced xhr with fetch
- Removed all xhr names from the docs
- Added note for users that they need to install a polyfill if they need to support old browsers
- The current API is still backward compatible

## What I didn't do (but I will do in future PRs)
- Remove xhr names from the public API (eg. xhrPath, xhrTimeout). I didn't do it right now since it would require users to update their code in order to benefit from this update;
- Fix the retry/abort bug that I mentioned on #278. I fixed it in a local branch but with xhr. Then I realized that I would have to rewrite everything again to make it work with fetch. So I opted to keep the bug for now;
- I didn't cleanup much the code. I tried to keep everything as it was as much as possible. While it was very tempting to refactor everything, I think the PR would get huge and hard to review. But the error handling part requires some love as well the unit tests.

Maybe the github diff is not that helpful here. But as I said, I tried to not change the code much. Mainly, the `io` function from `http.client` module had to be updated. The place that I think is super critical to review is the timeout implementation (so I  recommend you to triple check that).
 

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Universal data access layer for web applications.
 
-Typically on the server, you call your API or database directly to fetch some data. However, on the client, you cannot always call your services in the same way (i.e, cross domain policies). Instead, XHR requests need to be made to the server which get forwarded to your service.
+Typically on the server, you call your API or database directly to fetch some data. However, on the client, you cannot always call your services in the same way (i.e, cross domain policies). Instead, XHR/fetch requests need to be made to the server which get forwarded to your service.
 
 Having to write code differently for both environments is duplicative and error prone. Fetchr provides an abstraction layer over your data service calls so that you can fetch data using the same API on the server and client side.
 
@@ -149,8 +149,8 @@ See the [simple example](https://github.com/yahoo/fetchr/tree/master/examples/si
 
 ## Service Metadata
 
-Service calls on the client transparently become xhr requests.
-It is a good idea to set cache headers on common xhr calls.
+Service calls on the client transparently become fetch requests.
+It is a good idea to set cache headers on common fetch calls.
 You can do so by providing a third parameter in your service's callback.
 If you want to look at what headers were set by the service you just called,
 simply inspect the third parameter in the callback.
@@ -169,7 +169,7 @@ export default {
             headers: {
                 'cache-control': 'public, max-age=3600',
             },
-            statusCode: 200, // You can even provide a custom statusCode for the xhr response
+            statusCode: 200, // You can even provide a custom statusCode for the fetch response
         };
         callback(null, data, meta);
     },
@@ -241,9 +241,9 @@ fetcher
     });
 ```
 
-## XHR Object
+## Abort support
 
-The xhr object is returned by the `.end()` method as long as you're _not_ chaining promises.
+An object with an `abort` method is returned by the `.end()` method as long as you're _not_ chaining promises.
 This is useful if you want to abort a request before it is completed.
 
 ```js
@@ -253,11 +253,11 @@ const req = fetcher
     .end(function (err, data, meta) {
         // err.output will be { message: "Not found", more: "meta data" }
     });
-// req is the xhr object
+
 req.abort();
 ```
 
-However, you can't acces the xhr object if using promise chaining like so:
+However, due to the current implementation, you can't access this method if using promise chaining like so:
 
 ```js
 const req = fetcher
@@ -268,10 +268,10 @@ const req = fetcher
 req.then(onResolve, onReject);
 ```
 
-## XHR Timeouts
+## Timeouts
 
 `xhrTimeout` is an optional config property that allows you to set timeout (in ms) for all clientside requests, defaults to `3000`.
-On the clientside, xhrPath and xhrTimeout will be used for XHR requests.
+On the clientside, xhrPath and xhrTimeout will be used for all requests.
 On the serverside, xhrPath and xhrTimeout are not needed and are ignored.
 
 ```js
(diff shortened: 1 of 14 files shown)


pablopalacios:
@redonkulus I hope you haven't started reviewing it, I just realized that I sent the dirty branch.

pablopalacios:
Now it should be easier to review.

redonkulus:
Thanks for the update, looks like some conflicts after I merged your previous PR.

pablopalacios:
@redonkulus done ;)

redonkulus:
Sorry merged another of your PRs and created conflicts. One more rebase please.

pablopalacios:
@redonkulus no prob, done again :)

README.md
```diff
@@ -294,11 +301,20 @@ fetcher
     });
 ```
 
-## XHR Params Processing
+## Params Processing
 
-For some applications, there may be a situation where you need to process the service params passed in XHR request before params are sent to the actual service. Typically, you would process these params in the service itself. However, if you want to perform processing across many services (i.e. sanitization for security), then you can use the `paramsProcessor` option.
+For some applications, there may be a situation where you need to
+process the service params passed in the request before they are sent
```
redonkulus, line 307: Does your editor auto break the lines at a certain character count or is prettier doing this?

libs/util/http.client.js
```diff
@@ -174,62 +170,99 @@ function doXhr(method, url, headers, data, config, attempt, callback) {
...
+    }, options.timeout);
 
-                options.on.failure.call(null, err, resp);
+    fetch(request)
+        .then(function (response) {
+            clearTimeout(timeoutId);
+
+            if (response.ok) {
+                response.text().then(function (responseBody) {
+                    options.on.success(null, {
```
redonkulus, line 237: do we need to be defensive with options.on, is it always defined?

tests/functional/fetchr.test.js
```diff
@@ -163,13 +163,13 @@ describe('client/server integration', () => {
                 meta: { foo: 'bar' },
                 output: { message: 'error' },
                 rawRequest: {
-                    url: '/api/error?returnMeta=true',
+                    url: 'http://localhost:3000/api/error?returnMeta=true',
```
redonkulus, line 166: fetch require the full host now?

tests/unit/setup.js
```diff
@@ -0,0 +1,23 @@
...
+    AbortController,
+    abortableFetch,
+} = require('abortcontroller-polyfill/dist/cjs-ponyfill');
+
+const { fetch, Request } = abortableFetch({
+    fetch: nodeFetch,
+    Request: nodeFetch.Request,
+});
+
+if (!global.fetch) {
```
redonkulus, line 17: which envs are we testing in that do not support fetch?

README.md
pablopalacios: My editor. Should I undo it?

tests/functional/fetchr.test.js
pablopalacios: No. `rawRequest` here is `fetchr` custom serialization of a whatwg `Request` instance. `Request.url` returns the host and the path concatenated (it doesn't provide a `path` getter).

You can try on your browser:

```js
let req = new Request('/api')
console.log(req.url)
// https://github.com/api

req = new Request('https://yahoo.com/api')
console.log(req.url)
// https://yahoo.com/api
```

Also, since the field name is url, IMO it seems wrong to just set a path.

libs/util/http.client.js
redonkulus: ok

tests/functional/fetchr.test.js
redonkulus: ok good to know, thanks!

README.md
pablopalacios: Done ;)

lingyan (approved):
Code changes look good. @pablopalacios @redonkulus are we planning to bump major to 1.0, since apps will need to add fetch polyfill after this change?

pablopalacios:
@lingyan thanks for the review. This is a good question actually. We did some breaking changes when we moved from 0.5 to 0.6, so I would be fine moving to 0.7 now. 

If we go to 1.0, then changing the options API (#275) would require a bump to 2.0. But I would like to avoid many major bumps in a sequence. To avoid that I would close this PR and create a new one replacing `xhr` with native `XmlHttpRequest`. In this way it would still be possible to do some internal improvements before really changing things.

redonkulus:
I would not go to 1.x, saving that for the big api refactor. We just released 0.6 so I’m not too concerned about getting this out in a patch version. I doubt we have many users on this version anyways.

lingyan:
Theoretically we should bump to 0.7 or 1.0.  But if we know there are not other users using 0.6, I'm fine with bumping patch.

pablopalacios:
I would go with 0.7.

redonkulus:
Ok will merge and release 0.7 to be safe

tests/unit/setup.js
redonkulus: Good point 😄

redonkulus (approved):

redonkulus:
0.7 released https://github.com/yahoo/fetchr/releases/tag/v0.7.0

Thanks for all the hard work you are doing!

pablopalacios:
It's so nice how the package looks now on bundlephobia:
2021-09-29-185053_672x473_scrot
https://bundlephobia.com/package/fetchr@0.7.0

redonkulus:
wow ya its definitely trending downward

PR #293, opened 2021-09-26
refactor: remove deepmerge as dependency

pablopalacios:

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -10,7 +10,6 @@
  * @module Fetcher
  */
 var REST = require('./util/http.client');
-var deepmerge = require('deepmerge');
 var DEFAULT_GUID = 'g0';
 var DEFAULT_XHR_PATH = '/api';
 var DEFAULT_XHR_TIMEOUT = 3000;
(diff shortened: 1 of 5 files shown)


redonkulus:
Changing deepMerge for a shallow copy, you sure this won’t break any configs? Anything past the first level will not be merged.

pablopalacios:
Now I am.

I double checked what should be deep merged. To make it clear, in the client, `Fetchr` can be initiated with the following options:

- `context`: a flat object where the values must be strings (since they are passed to `encodeURIComponent` later)
- `contextPicker`: an object that the values can be a string, an array or a function
- `corsPath`: a string
- `headers`: a flat object where the values must be strings (since they become headers values later)
- `statsCollector`: a function
- `xhrPath`: a string
- `xhrTimeout`: a number

So, `corsPath`, `statsCollector`, `xhrPath`, `xhrTimeout` have no need to be deep merged. `context` and `headers` can be individually shallowly copied since their values must be a string. 

`contextPicker`, however, requires attention since its values can be a string, a function or an array of strings. Because of this last case, we can't just replace the values. If both old and new values are arrays, we must merge them as well.

I added the missing code to handle `contextPicker` properly (and a bunch of tests for all update cases) and, unless I'm missing some case, it should be good to go.

One thing to mention though. Since 2015, in the server side, update options just do a shallow copy, it's only in the client that the deepmerge happened. I'm pretty sure that we should be consistent here.

redonkulus (approved):
Looks good to me, thank you for the detailed info and adding the extra tests!

PR #335, opened 2021-12-14
Remove support for returnMeta query param

pablopalacios:
`returnMeta` query param was introduced in order to handle a response format update from the server. There was a time that we would return the data from the resources without making any change. When `returnMeta` was introduced, we started formatting the response as following:

```js
// successful responses:
{
    data: serviceResponseData,
    meta: serviceResponseMeta,
}

// error responses:
{
    output: serviceErrorOutput,
    meta: serviceResponseMeta,
}
```
`returnMeta` was then added to prevent outdated clients to break because of the new format. Years have passed and all browsers should have a fresh copy of fetchr by now. Therefore, we can remove the support for it as requested in the TODO comments (ex. https://github.com/yahoo/fetchr/blob/241a71c204b17f25f7c5ff403e054ab70782425d/libs/fetcher.client.js#L272).

One caveat though: when the new format was added, we forgot to update the POST error responses. This PR also addresses this issue.

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -459,31 +459,21 @@ Fetcher.middleware = function (options) {
                     }
                     if (err) {
                         var errResponse = getErrorResponse(err);
-                        if (req.query && req.query.returnMeta) {
-                            res.status(errResponse.statusCode).json(
-                                responseFormatter(req, res, {
-                                    output: errResponse.output,
-                                    meta: meta,
-                                })
-                            );
-                        } else {
-                            res.status(errResponse.statusCode).json(
-                                responseFormatter(req, res, errResponse.output)
-                            );
-                        }
-                        return;
-                    }
-                    if (req.query.returnMeta) {
-                        res.status(meta.statusCode || 200).json(
+                        res.status(errResponse.statusCode).json(
                             responseFormatter(req, res, {
-                                data: data,
+                                output: errResponse.output,
                                 meta: meta,
                             })
                         );
-                    } else {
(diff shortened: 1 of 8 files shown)


libs/fetcher.js
```diff
@@ -459,31 +459,21 @@ Fetcher.middleware = function (options) {
...
-                        // TODO: Remove `returnMeta` feature flag after next release
-                        res.status(meta.statusCode || 200).json(data);
+                        return;
                     }
+
+                    res.status(meta.statusCode || 200).json(
+                        responseFormatter(req, res, {
+                            data: data,
+                            meta: meta,
+                        })
```
lingyan, line 475: thanks for cleaning this TODO item up, @pablopalacios !
@redonkulus and I were talking about this recently, about why GET success case wasn't adding the meta data :)

@redonkulus, I checked our internal code using responseFormatter. this changes seems to be ok.  You wanna double check?

redonkulus: We are already on v0.7.0 publicly, internally we are on fluxible-plugin-fetch@0.3.x which uses fetchr@0.5.x so this change will not affect us.

pablopalacios: @lingyan oh, but from what I see in the code, only the POST failed  responses were not including the meta data. Unless you use a very old fetchr version, I think you should already have meta available. Could it be that the query parser or any other middleware is modifying the query object?

redonkulus: We are in fact on an old version of fetchr, we have not upgraded to the latest (as it requires regression testing across many apps).

PR #336, opened 2021-12-14
client: remove X-Requested-With request header

pablopalacios:
We don't use XMLHttpRequest anymore.

### DIFF
--- a/libs/util/http.client.js
+++ b/libs/util/http.client.js
@@ -31,11 +31,8 @@ var DEFAULT_CONFIG = {
 
 var INITIAL_ATTEMPT = 0;
 
-function normalizeHeaders(headers, method, isCors) {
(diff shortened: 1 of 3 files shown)


redonkulus:
@pablopalacios need to rebase because I merged your other PR. Also, remind me again why we removed this header?

pablopalacios:
@redonkulus we use native fetch instead of XMLHttpRequest, so it would make sense to update the value of the header. But we are not consuming this header at all, so I thought it would be fine to remove it. Do you use it for any reason?

redonkulus:
Yes, I believe there may be some code internally that uses this header to differentiate between a client and server request to a service. I think we can leave it in, the value is a bit of a misnomer. Perhaps would've been better to have a value of `client` instead.

PR #338, opened 2021-12-16
feat: add resource name in POST path requests

pablopalacios:
Currently, we only add the resource name in the path for GET requests:

`fetchr.read('resource-name')` => `/api/resource-name`.

But we never do it for POST requests:

`fetchr.read('resource-name').params(hugePayload)` => `/api`.

This behavior makes it hard to build monitoring tools on top of it. For example, one could have a gateway in front of `fetchr` that monitors the response success rate per path and triggers an alert if the rate drops. For GET requests, it is easy to spot which resource is the faulty one, but for POST requests it's just impossible without diving into the logs of the service where fetchr is running. It's definitely doable, but not easy if the team that manages the gateway have no idea what is behind it.

Another challenge would be to add a circuit breaker per resource. Currently we would need to parse the whole request to implement it (which requires knowledge about how fetchr works internally). But if we always have the resource in the path, then it would be way easier.

In this PR, I just did the least effort change: add the resource in the path if user haven't provided a custom path. But if you think the general idea is valid but the implementation could be different, as always, I'm open to change it :)

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -309,7 +309,7 @@ function executeRequest(request, resolve, reject) {
         requests: requests,
         context: request.options.context,
     }; // TODO: remove. leave here for now for backward compatibility
-    uri = request._constructGroupUri(uri);
+    uri = request._constructPostUri(uri);
     allow_retry_post = request.operation === OP_READ;
     return REST.post(
         uri,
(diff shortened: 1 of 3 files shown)


redonkulus (approved):
Looks good, thanks!

PR #342, opened 2021-12-24
refactor: use modern javascript in the server code

pablopalacios:
This changes should even work with the already unsupported Node 10 (the functional tests pass, but unit don't since we already use some modern APIs there).

### DIFF
--- a/libs/fetcher.js
+++ b/libs/fetcher.js
@@ -82,127 +82,129 @@ function getErrorResponse(err) {
  *      the req object, the serviceInfo object, and the params object.  It is expected to return the processed params object.
  * @constructor
  */
-function Request(operation, resource, options = {}) {
-    if (!resource) {
-        throw new Error('Resource is required for a fetcher request');
-    }
-
-    this.operation = operation || OP_READ;
-    this.resource = resource;
-    this.req = options.req || {};
-    this.serviceMeta = options.serviceMeta || [];
-    this._params = {};
-    this._body = null;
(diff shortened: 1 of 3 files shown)


libs/fetcher.js
```diff
@@ -273,8 +259,8 @@ function executeRequest(request, resolve, reject) {
  *      the req object, the serviceInfo object, and the params object.  It is expected to return the processed params object.
  * @constructor
  */
-function Fetcher(options) {
-    this.options = options || {};
+function Fetcher(options = {}) {
```
redonkulus, line 262: should we just make this a class now?

libs/fetcher.js
```diff
@@ -81,14 +82,13 @@ function getErrorResponse(err) {
  *      the req object, the serviceInfo object, and the params object.  It is expected to return the processed params object.
  * @constructor
  */
-function Request(operation, resource, options) {
+function Request(operation, resource, options = {}) {
```
redonkulus, line 85: this could be a class too i think

libs/fetcher.js
pablopalacios: It can be. I didn't change at first because of the size of the diff. But since you pointed, I converted to modern class.

libs/fetcher.js
pablopalacios: Same as the other one :)

redonkulus (approved):
Looks great, thanks for the refactor!

PR #110, opened 2015-07-13
[Fixes #108] Refactor CRUD interface

Vijar:
@redonkulus @mridgway @lingyan 

Sorry this is so huge, one thing led to another down the rabbit hole. This PR includes:
- New interface for CRUD methods w/ backwards compatibility
    - Deprecated old interface
- Renamed all instances of 'fetchers' to 'services' w/ backwards compatibility
    - Deprecated `registerFetcher` and `getFetcher`
- Fixed an issue where custom `constructGetUri` function wasn't falling back properly
- Housekeeping on test suite
- Deleted old batching code

PR #343, opened 2021-12-27
refactor: remove unused batching code and remove duplication

pablopalacios:
Support for batched requests was drop a long time ago (see: https://github.com/yahoo/fetchr/pull/110) but we still have some code from that time.

By removing it completely, GET and POST requests are almost identical (the only difference is that POST has a body). This allows us to do some refactoring to remove many code duplication.

The change is quite big (and is on top of #342, so I recommend to merge that one first) but hopefully, reviewing commit by commit should make it easier to review.

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -17,10 +17,6 @@ var OP_READ = 'read';
 var defaultConstructGetUri = require('./util/defaultConstructGetUri');
 var forEach = require('./util/forEach');
 
-function isFunction(value) {
-    return typeof value === 'function';
-}
-
 function parseResponse(response) {
     if (response && response.responseText) {
         try {
@@ -248,9 +244,10 @@ function executeRequest(request, resolve, reject) {
     }
 
     if (request.operation === OP_READ) {
-        var buildGetUrl = isFunction(config.constructGetUri)
-            ? config.constructGetUri
-            : defaultConstructGetUri;
+        var buildGetUrl =
+            typeof config.constructGetUri === 'function'
+                ? config.constructGetUri
+                : defaultConstructGetUri;
 
         var context = pickContext(
             request.options.context,
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -223,106 +223,70 @@ Request.prototype.end = function (callback) {
  * @param {Function} reject function to call when request rejected
  */
 function executeRequest(request, resolve, reject) {
-    var clientConfig = request._clientConfig;
-    var use_post;
-    var allow_retry_post;
-    var uri = clientConfig.uri;
-    var data;
+    var callback = function (err, response) {
+        if (err) {
+            return reject(err);
+        }
+        resolve(parseResponse(response));
+    };
 
-    if (!uri) {
-        uri = clientConfig.cors
+    var config = Object.assign(
+        {
+            unsafeAllowRetry: request.operation === OP_READ,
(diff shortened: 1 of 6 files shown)


libs/fetcher.js
```diff
@@ -49,9 +48,7 @@ function parseRequest(req) {
 }
 
 function sanitizeResourceName(resource) {
-    return resource
-        ? resource.replace(RESOURCE_SANTIZER_REGEXP, '*')
-        : resource;
+    return resource ? resource.replace(/[^\w.]+/g, '*') : resource;
```
redonkulus, line 51: Was this to avoid a bug? We usually keep the regex out to avoid recreating the regex object each request

redonkulus:
I went commit by commit and it looks pretty good to me, you can rebase since the other PR was merged. Then I can leave any feedback on removed badge code changes.

libs/fetcher.js
pablopalacios: Oh, I didn't know that. I'll remove this commit.

libs/fetcher.client.js
```diff
@@ -10,65 +10,14 @@
  * @module Fetcher
  */
 var REST = require('./util/http.client');
-var DEFAULT_GUID = 'g0';
+var defaultConstructGetUri = require('./util/defaultConstructGetUri');
```
redonkulus, line 13: looks like this file can move to modern code too, for another PR

libs/util/http.client.js
```diff
@@ -26,6 +26,17 @@ var DEFAULT_CONFIG = {
 
 var INITIAL_ATTEMPT = 0;
 
+function parseResponse(response) {
```
redonkulus, line 29: this was in fetcher.client.js but not needed there?

libs/fetcher.client.js
pablopalacios: Oh, only if we completely drop support for IE11 (or add ts or babel to transpile correctly).

libs/fetcher.js
```diff
@@ -36,12 +35,55 @@ function parseParamValues(params) {
     }, {});
 }
 
+function parseRequest(req) {
```
redonkulus, line 38: some of these functions could probably move to their own file as well to reduce the fetcher.js size

redonkulus (approved):

libs/fetcher.client.js
redonkulus: Yup, I created #344  to convert to TS

redonkulus:
Released in 0.7.2

libs/util/http.client.js
pablopalacios: It is not that it's not needed. But it's inconsistent that we parse error responses inside the http module but normal responses we parse inside the fetcher module. Now it's everything in one place (and we can potentially use `fetch().(response => response.json()))` instead of `text()` (see: https://github.com/yahoo/fetchr/blob/80923dc7d908aae547e0a8b60528fecfc0943177/libs/util/http.client.js#L223). This change was not really related to this PR but it was lingering in my machine for an while.

libs/fetcher.js
pablopalacios: Perhaps with the TS migration we can re-organize the folders. It's clear to me that we have actually to packages here, one for the server, another for the client. It seems wrong to have everything in the same folder as we have right now. So I'm avoiding to split the server code until we don't have things better organized (by the way, `utils` has only client side code now).

PR #347, opened 2022-01-05
refactor(http): remove callbacks, simplify options normalization and export only one function

pablopalacios:
Summary of changes:
- replace `get` and `post` with only one method:`httpRequest`
- `httpRequest` always returns a promise (fetcher is now the only responsible for handling the callback flow)
- Simplify http request options normalization

Again, I recommend you to review commit by commit ;)

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -9,7 +9,7 @@
  * Fetcher is a CRUD interface for your data.
  * @module Fetcher
  */
-var REST = require('./util/http.client');
+var httpRequest = require('./util/http.client').default;
 var defaultConstructGetUri = require('./util/defaultConstructGetUri');
 var forEach = require('./util/forEach');
 var pickContext = require('./util/pickContext');
@@ -172,6 +172,8 @@ Request.prototype.end = function (callback) {
  * @param {Function} reject function to call when request rejected
  */
 function executeRequest(request) {
+    var options = {};
+
     var config = Object.assign(
         {
             unsafeAllowRetry: request.operation === OP_READ,
@@ -179,9 +181,9 @@ function executeRequest(request) {
         },
         request._clientConfig
     );
-    var headers = config.headers || request.options.headers || {};
+    options.config = config;
+    options.headers = config.headers || request.options.headers || {};
 
-    var url;
     var baseUrl = config.uri;
     if (!baseUrl) {
         baseUrl = config.cors
@@ -190,6 +192,8 @@ function executeRequest(request) {
     }
 
     if (request.operation === OP_READ && !config.post_for_read) {
+        options.method = 'GET';
+
         var buildGetUrl =
             typeof config.constructGetUri === 'function'
                 ? config.constructGetUri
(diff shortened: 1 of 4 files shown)


redonkulus:
@pablopalacios is this part of the 1.0 breaking changes / api refactor work?

redonkulus:
Looking at the code, I wonder if we should just remove all the .then() calls and switch to async / await everywhere. Especially if we are removing callbacks. Its much easier to maintain and read. Then we can just document that this library only works with latest browsers and node versions.

libs/fetcher.client.js
```diff
@@ -128,11 +128,11 @@ Request.prototype._captureMetaAndStats = function (err, result) {
 Request.prototype.end = function (callback) {
```
redonkulus, line 128: This callback is still required?

libs/fetcher.client.js
```diff
@@ -187,17 +181,19 @@ function executeRequest(request, resolve, reject) {
...
-    var url;
     var baseUrl = config.uri;
     if (!baseUrl) {
         baseUrl = config.cors
             ? request.options.corsPath
             : request.options.xhrPath;
     }
 
-    if (request.operation === OP_READ) {
+    if (request.operation === OP_READ && !config.post_for_read) {
```
redonkulus, line 194: post for read is weird, I know this is old, what is the use case for this again?

pablopalacios: To be really honest, I don't know. I see in my code base that we used to use `create` instead of `read` just to make sure we would have a POST. I think the reason for that was related to how our bot protection used to work. In that case, calling `read` with `post_for_read` would be semantically better.

libs/util/http.client.js
```diff
@@ -200,111 +135,81 @@ function FetchrError(options, request, response, responseBody, originalError) {
...
-            url,
-            headers,
-            data,
-            config,
-            INITIAL_ATTEMPT,
-            callback
-        );
-    },
-};
+module.exports.default = httpRequest;
```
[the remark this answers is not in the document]
pablopalacios: No big reason behind it. I names as default since probably in the future, this line of code will become: `export default httpRequest`. 

How would you prefer? We can export it as `httpRequest` (or any other name). The only thing that I'm afraid is to export it directly (`module.exports = httpRequest`) since it might break the tests because of the mocks.

libs/fetcher.client.js
pablopalacios: By the way, I checked that the support exists since the very first implementation. Don't you have any usage of it in your code base?

pablopalacios:
> @pablopalacios is this part of the 1.0 breaking changes / api refactor work?

Actually no (but it definitely will help). I've found some corner cases that are not being covered regarding aborts, timeouts and retries. But in order to fix it, the http module must be simplified first. So before any breaking change, I would like to tidy up the internals and have http working 100%.

> Looking at the code, I wonder if we should just remove all the .then() calls and switch to async / await everywhere. Especially if we are removing callbacks. Its much easier to maintain and read. Then we can just document that this library only works with latest browsers and node versions.

I'm not so sure if we can fully migrate to async/await. Since we have abort as part of the API, we must return something to the user before getting the response. But I think after tidying up the code, it will get easier to understand (even though I think doing a fetch-abort-retry-timeout will always end up in a not so straightforward code).

libs/util/http.client.js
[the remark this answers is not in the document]
redonkulus: ok we can leave for now, once if/when move to TS, these will all becomes import/export anyways

redonkulus (approved):
Changes look good to me, good to see more consolidation of code

PR #348, opened 2022-01-11
refactor: join options logic from http and fetchr client modules

pablopalacios:
This is the last big refactoring one. Here, I'm merging the code related to options normalization from the `http.client` and `fetchr.cleint` modules. As you can see, when you put both together, we have a bunch of code (that I'll probably simplify in another PR).

By centralizing the options we basically create only one interface that is passed down to all functions (it will be helpful when migrating to ts). 

But more importantly, `httpRequest` does not compute the options unnecessarily on every retry. This makes the plate clean to fix the corner cases that I mentioned in the other PR:
- do not retry user requests aborted by the user
- add abort support for the promise flow

There will be also another PR improving the `FetchrError` class:
- Make it inherit from `Error`;
- Attach the `options` object;
- Add `name`, `message` and `reason` fields to make it easier for users to report/log errors;

Since all this changes should be backward compatible, I would like to do it before migrating to ts/change the API.

### DIFF
--- a/libs/util/httpRequest.js
+++ b/libs/util/httpRequest.js
@@ -4,7 +4,7 @@
  */
 
 /**
- * @module rest-http
+ * @module httpRequest
  */
 
 function FetchrError(options, request, response, responseBody, originalError) {
@@ -138,4 +138,4 @@ function httpRequest(options, attempt) {
     };
 }
 
-module.exports.default = httpRequest;
+module.exports = httpRequest;
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -14,7 +14,6 @@ var normalizeOptions = require('./util/normalizeOptions');
 
 var DEFAULT_PATH = '/api';
 var DEFAULT_TIMEOUT = 3000;
-var OP_READ = 'read';
 
 /**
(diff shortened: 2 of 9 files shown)


libs/util/httpRequest.js
```diff
@@ -135,26 +50,46 @@ function FetchrError(options, request, response, responseBody, originalError) {
...
 
     return fetch(request).then(
         function (response) {
             clearTimeout(timeoutId);
 
             if (response.ok) {
-                return response.text().then(function (responseBody) {
-                    return parseResponse(responseBody);
+                return response.json().catch(function () {
+                    return null;
```
redonkulus, line 92: return null?

redonkulus (approved):
Overall changes look good to me

libs/util/httpRequest.js
pablopalacios: Yup :( This is the fetchr behavior since a looong time ago. For me, we could throw an error in this case as well, but I'm not sure yet if we would be fixing or introducing a bug. 

I need to add some functional tests to properly see how fetchr will behave with a 200 response and empty body. I think unless fetchr is not properly set in the server, there is no way to get a empty response...

PR #352, opened 2022-01-12
refactor: simplify FetchrError and fix Error inheritance

pablopalacios:
After migrating from xhr to native fetch, many cases inside `FetchrError` were not reachable anymore. I didn't refactor it back that time since the code was super hard to understand and we didn't have functional tests to see the real consequences of updating it. But now it was a breeze.

Hopefully, FetchrError class is easier to understand now. I also briefly document it in the README and, as you can see, the data inside the error could be better and simplified. But it would introduce some breaking changes that I would postpone for V1 now.

### DIFF
--- a/libs/util/httpRequest.js
+++ b/libs/util/httpRequest.js
@@ -13,21 +13,15 @@ function FetchrError(options, request, response, responseBody, originalError) {
     var errMessage, errBody;
 
     if (!err && (status === 0 || (status >= 400 && status < 600))) {
-        if (typeof responseBody === 'string') {
-            try {
-                errBody = JSON.parse(responseBody);
-                if (errBody.message) {
-                    errMessage = errBody.message;
-                } else {
-                    errMessage = responseBody;
-                }
-            } catch (e) {
+        try {
+            errBody = JSON.parse(responseBody);
+            if (errBody.message) {
+                errMessage = errBody.message;
+            } else {
                 errMessage = responseBody;
             }
-        } else {
-            errMessage = status
-                ? 'Error ' + status
-                : 'Internal Fetchr XMLHttpRequest Error';
+        } catch (e) {
+            errMessage = responseBody;
         }
 
         err = new Error(errMessage);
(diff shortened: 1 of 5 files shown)


libs/util/httpRequest.js
```diff
@@ -92,19 +79,14 @@ function io(options, controller) {
...
                 });
             } else {
-                return response.text().then(function (responseBody) {
-                    throw new FetchrError(
-                        options,
-                        request,
-                        response,
-                        responseBody
-                    );
+                return response.text().then(function (body) {
```
redonkulus, line 82: is body an error here? should we call it error instead?

libs/util/httpRequest.js
```diff
@@ -7,49 +7,36 @@
...
+            this.output = this.body.output || null;
+            this.meta = this.body.meta || null;
+            this.message = this.body.message || message;
+        } catch (e) {
+            this.body = message;
+        }
+    }
 }
 
+FetchrError.prototype = Object.create(Error.prototype);
```
redonkulus, line 37: Why not make it a proper class?

libs/util/httpRequest.js
pablopalacios: Do you mean an `Error` instance or that the body contains an error message? Well, it's a string and it might contain some error message yes.
But I think it makes sense what you said, however, I think renaming to `err` could be misleading since, normally, `err` points to an `Error` instance. It gets more awkward since just a few line below, we have another `err` name, and that one is a true error.
So I think renaming to `message` could be a good option (since this is the name of `FetchrError` param as well). What do you think?

libs/util/httpRequest.js
pablopalacios: To keep es5 compatibility. I will transform it as soon as we move to TS, but for now, I think it's better to keep es5 syntax compatibility.

libs/util/httpRequest.js
redonkulus: message makes sense to me

redonkulus (approved):
LGTM, thanks!

PR #359, opened 2022-01-21
Improve request abort, retry and error handling

pablopalacios:
@redonkulus now I would say that the httRequest module is 100%. 

- Add `reason` property in `FetchrError`
- Differentiate timeout errors from abort errors
- Add abort support for promise requests
- Do not retry requests aborted by the user
- Do not accept non JSON responses


pablopalacios:
By the way, we have a cool dashboard for fetchr errors in our company: 

2022-01-21-230611_1877x910_scrot

I'm pretty sure that you guys have way more cool stuff on top of fetchr and you probably handle way more requests than us, but I just wanted to point that, even though we changed fetchr a lot, it's been running quite successfully for us.

redonkulus:
This is great, ya we splunk too and some internal tools to build dashboards for our API requests. But this looks pretty good!

libs/util/httpRequest.js
```diff
@@ -7,7 +7,7 @@
  * @module httpRequest
  */
 
-function FetchrError(message, options, request, response) {
+function FetchrError(reason, message, options, request, response) {
```
redonkulus, line 10: probably one more argument away from just passing them all in as an object

libs/util/httpRequest.js
```diff
@@ -76,17 +83,42 @@ function io(options, controller) {
 
             if (response.ok) {
                 return response.json().catch(function () {
-                    return null;
+                    throw new FetchrError(
+                        'BAD_JSON',
```
redonkulus, line 87: I'm thinking it might be nice to move these error codes into a constants file and export them, that way we can also use them in the tests and not have to worry about typo errors in the future.

libs/util/httpRequest.js
pablopalacios: Yup, but it's a pain in the ass to work with objects as arguments in es5. I'll leave it for now and refactor when we move to ts.

libs/util/httpRequest.js
pablopalacios: I moved FetchrError to a separated module and set the reasons as static proprties in the class. I'm not sure if it wouldn't be better to have a separate class per kind of error instead of the reason field. But again, because of es5, doing inheritance is a pain in the ass...

redonkulus: Yup makes sense to keep it ES5 compat for now. Changes look good.

redonkulus (approved):

redonkulus:
Released https://github.com/yahoo/fetchr/releases/tag/v0.7.7

PR #365, opened 2022-01-26
fix: allow requests to be aborted after retries and timeouts

pablopalacios:
If we don't create a new AbortController on every retry, the abort
method will have a reference to the first controller that is already
expired.

We must also pay attention to not use the same controller on every
request, otherwise retries would not be possible.

### DIFF
--- a/libs/util/httpRequest.js
+++ b/libs/util/httpRequest.js
@@ -104,11 +104,11 @@ function io(options, controller) {
     );
 }
 
-function httpRequest(options, attempt) {
+function httpRequest(options) {
     var controller = new AbortController();
-    var currentAttempt = attempt || 0;
+    var currentAttempt = 0;
 
-    var promise = io(options, controller).catch(function (err) {
+    var promise = io(options, controller).catch(function retry(err) {
         if (!shouldRetry(err, options, currentAttempt)) {
             throw err;
         }
@@ -121,15 +121,19 @@ function httpRequest(options, attempt) {
             options.retry.interval *
             Math.pow(2, currentAttempt);
 
+        controller = new AbortController();
+        currentAttempt += 1;
         return delayPromise(function () {
-            return httpRequest(options, currentAttempt + 1);
+            return io(options, controller).catch(retry);
         }, delay);
     });
 
     return {
         then: promise.then.bind(promise),
         catch: promise.catch.bind(promise),
-        abort: controller.abort.bind(controller),
+        abort: function () {
+            return controller.abort();
+        },
     };
 }
 
--- a/libs/util/httpRequest.js
+++ b/libs/util/httpRequest.js
@@ -108,7 +108,20 @@ function httpRequest(options) {
     var controller = new AbortController();
     var currentAttempt = 0;
 
-    var promise = io(options, controller).catch(function retry(err) {
+    // handleError is the onReject promise callback that we attach to
+    // the io call (ex. io().catch(handleError)). Since io is a
+    // promise and since we must be able to retry requests (aka call
+    // io function again), we must call io from within
+    // handleError. This means that handleError is a recursive
+    // function. Recursive promises are problematic since they can
+    // block the main thread for a while. However, since the inner io
+    // call is wrapped in a setTimeout (through delayPromise) we are
+    // safe here.
+    //
+    // The call flow:
+    //
+    // httpRequest -> io -> handleError -> io -> handleError -> end
+    function handleError(err) {
         if (!shouldRetry(err, options, currentAttempt)) {
             throw err;
         }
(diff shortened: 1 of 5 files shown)


pablopalacios:
This one is super tricky, take a beer before reviewing it.

tests/functional/app.js
```diff
@@ -5,10 +5,13 @@ const { itemsService } = require('./resources/item');
 const { errorsService } = require('./resources/error');
 const { headersService } = require('./resources/headers');
 const { slowService } = require('./resources/slow');
+const { slowFastService } = require('./resources/slowFast');
```
redonkulus, line 8: slow fast?

pablopalacios: Hmmm, I should document/name it better. But the idea of this one is to give 2 slow responses and then a fast one. In this way it's possible to force the client to cancel the first 2 requests due to timeout but still get a good response in the 3rd try. Here, we are making sure that we are replacing the AbortController correctly between retries (instead of using the same one that would trigger an already aborted fetch call).

libs/util/httpRequest.js
```diff
@@ -121,15 +121,19 @@ function httpRequest(options, attempt) {
             options.retry.interval *
             Math.pow(2, currentAttempt);
 
+        controller = new AbortController();
+        currentAttempt += 1;
         return delayPromise(function () {
-            return httpRequest(options, currentAttempt + 1);
+            return io(options, controller).catch(retry);
```
pablopalacios, line 127: @redonkulus the important change is: `httpRequest` became a closure instead of a recursive function. However, note that the `catch` handler is recursive (and it's only possible since we named the handler). `retry` updates the `currentAttempt` and the `controller` and then triggers a new `io` call.

Recursive promises are dangerous since they can block the loop for an while, but since `delayPromise` calls `setTimeout` internally, we are safe here.

libs/util/httpRequest.js
```diff
@@ -121,15 +121,19 @@ function httpRequest(options, attempt) {
...
         }, delay);
     });
 
     return {
         then: promise.then.bind(promise),
         catch: promise.catch.bind(promise),
-        abort: controller.abort.bind(controller),
+        abort: function () {
+            return controller.abort();
+        },
```
pablopalacios, line 136: @redonkulus it was necessary to create a proper function since it might be that the initial `controller` is already stale. So, here, whenever users calls `abort` we are always using a fresh reference.

tests/functional/fetchr.test.js
```diff
@@ -309,6 +328,22 @@ describe('client/server integration', () => {
...
+                        .then(() =>
+                            fetcher.read('slow-then-fast', null, {
+                                retry: { maxRetries: 5 },
+                                timeout: 80,
+                            })
+                        );
+                });
+
+                expect(response.data.attempts).to.equal(3);
+            });
```
pablopalacios, line 346: @redonkulus This test case was added mainly to make sure that I was not introducing any regression.

tests/functional/fetchr.test.js
```diff
@@ -266,6 +266,25 @@ describe('client/server integration', () => {
...
+                    return new Promise((resolve) =>
+                        setTimeout(() => {
+                            promise.abort();
+                            resolve();
+                        }, 500)
+                    ).then(() => promise.catch((err) => err));
+                });
+
+                expect(response.reason).to.equal(FetchrError.ABORT);
+            });
```
pablopalacios, line 286: @redonkulus this was the test that was failing with the previous implementation. It's ugly, but, basically, it triggers a call to the slow resource (which always takes 5s to respond) with a timeout of 200ms. After that, we schedule an `abort` call after 500ms (in the middle of the 3rd call).

This test would fail since we would get a `TIMEOUT` error (meaning that we were aborting a controller that was already stale). But now, the correct error is sent (meaning that we were able to prevent the last 2 calls to happen).

pablopalacios:
@redonkulus before making the changes here, I actually created a small repo trying to cover all the corner-cases when handling abort, timeout and retries with fetch: https://github.com/pablopalacios/fetch-art

libs/util/httpRequest.js
redonkulus: lets add this comment in the code for future reference

libs/util/httpRequest.js
redonkulus: would be good to add description as a comment in the code

tests/functional/fetchr.test.js
redonkulus: again, i like more details in the code than not, feel free to add this scenario as a comment too

redonkulus (approved):
looks good to me, thank you for adding the comments, it will help later when i forget all this

PR #380, opened 2022-04-02
feat(client): add support for global retry configuration

pablopalacios:
@redonkulus we were only able to set retry configuration per request. Now, it's possible to set some (or all) configuration globally:

```js
const fetchr = new Fetchr({
    retry: {
     interval: 350,
     statusCodes: [0, 502, 504],
   },
});

fetchr.read('resource').clientConfig({ retry: { maxRetries: 2 } }).end();
```

This avoids repeating some configuration that (at least in our case) are always the same.

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -184,7 +184,9 @@ function Fetcher(options) {
         corsPath: options.corsPath,
         context: options.context || {},
         contextPicker: options.contextPicker || {},
+        retry: options.retry || null,
         statsCollector: options.statsCollector,
+        unsafeAllowRetry: Boolean(options.unsafeAllowRetry),
         _serviceMeta: this._serviceMeta,
     };
 }
--- a/README.md
+++ b/README.md
@@ -433,21 +433,43 @@ For requests from the server, the config object is simply passed into the servic
 
 ## Retry
 
-You can set Fetchr to retry failed requests automatically by setting a `retry` settings in the client configuration:
+You can set Fetchr to automatically retry failed requests by specifying a `retry` configuration in the global or in the request configuration:
 
 ```js
-fetcher
+// Globally
+const fetchr = new Fetchr({
+    retry: { maxRetries: 2 },
+});
+
+// Per request
+fetchr
     .read('service')
     .clientConfig({
-        retry: {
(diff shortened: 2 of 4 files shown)


pablopalacios:
Soooo, @redonkulus , would it be possible to check this one?

libs/fetcher.client.js
```diff
@@ -184,7 +184,9 @@ function Fetcher(options) {
         corsPath: options.corsPath,
         context: options.context || {},
         contextPicker: options.contextPicker || {},
+        retry: options.retry || null,
         statsCollector: options.statsCollector,
+        unsafeAllowRetry: Boolean(options.unsafeAllowRetry),
```
redonkulus, line 189: Why is called unsafe?

redonkulus (commented):
I’m fine with a global config. Added one comment. We should document this in the readme too.

libs/fetcher.client.js
pablopalacios: the `unsafeAllowRetry` allows POST requests to be retried. In many cases, however, one would not like to have this feature enabled (since it could lead to duplicated transactions in the back-end). For me, a better name would be `retryOnPost` or something like this so it's clear what is happening (or, perhaps, removing this flag completely since, right now, retry is disabled by default).

Regarding the documentation, it's already in the readme: https://github.com/yahoo/fetchr/blob/081b3c7018d5444b6a35b65e6029b260f97075f0/README.md?plain=1#L477

konarkmodi:
@redonkulus @pablopalacios : Thanks to both of you for keeping this project alive and kicking. It's total bliss to use it.
Would it be possible to move this PR ahead and make the retry options globally available?

redonkulus:
@konarkmodi thanks for the kind words. @pablopalacios do you want to update the readme, then we can get this merged?

pablopalacios:
@redonkulus I thought it was already there. In any case, I updated the documentation to make it clearer (hopefully :D).

redonkulus (approved):
Thanks for the docs update, its clearer now.

redonkulus:
https://github.com/yahoo/fetchr/releases/tag/v0.7.9

PR #437, opened 2023-08-30
tests: always use node-fetch to simulate browser fetch

pablopalacios:
Since `httpRequest` is only used in the browser, it doesn't matter much if we are using `node-fetch`, `isomorphic-fetch`, `undici` or node's native fetch. Said that, to avoid tests to fail on Node 18 (which includes a native fetch implementation), we can overwrite the global fetch with node-fetch's implementation.

### DIFF
--- a/tests/unit/setup.js
+++ b/tests/unit/setup.js
@@ -14,10 +14,8 @@ const { fetch, Request } = abortableFetch({
     Request: nodeFetch.Request,
 });
 
-if (!global.fetch) {
-    global.AbortController = AbortController;
-    global.Headers = nodeFetch.Headers;
-    global.Request = Request;
-    global.Response = nodeFetch.Response;
-    global.fetch = fetch;
-}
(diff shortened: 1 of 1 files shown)


pablopalacios:
By the way, the tests seem to fail because of an undici issue: https://github.com/cloudflare/miniflare/issues/454. In the browser, `new Request('/api')` works fine, but in Node 18, it throws an invalid URL error since the host is missing. According to fetch spec, the browser behavior is correct, but it's also understandable that it wouldn't work in Node since there is no base uri set.
2023-08-30-233540_599x166_scrot

redonkulus (approved):
Thank you for the quick turnaround!

Interesting fix, makes sense to me. I was looking into the Request not having the host but was getting tripped up since there wasn't one in Node. It would be nice long term to use the built in fetch but I understand for testing a client only library, that this is necessary.

PR #514, opened 2024-07-16
fix: make options optional in fetchr client just as in server

pablopalacios:
In the server Fetchr, the options passed to the constructor are optional and we then use default values.

That was not the same in the client Fetchr.


pablopalacios:
@redonkulus I'm back :D

redonkulus (approved):

PR #515, opened 2024-07-21
feat: make Request object thenable

pablopalacios:
Now the request object returned by fetchr crud methods are [thenable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) by default. 

To make it possible, I had to refactor `httpRequest` to transform it in a class to make it easier to expose `then`, `catch` and `abort` to the users. By doing it, it was also possible to simplify the `Request` class and remove some workarounds related to `abort`.

This PR also implement some of the stuff we discussed in https://github.com/yahoo/fetchr/pull/263#issuecomment-881089203 without the need for a major version bump. 

## Before
```js
// superagent style
fetchr
  .read('foo')
  .end()
  .then()

// callback style without callback
fetchr
  .read('foo', null) // passing a second params was necessary to not trigger the superagent API.
  .then() 
```
## After
```js
// superagent or callback
fetchr
  .read('foo')
  .then()
```

This is also quite nice when doing fire and forget calls with async/await:

```js
await fetchr.update('counter');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,19 +106,22 @@ Fetcher.registerService(myDataService);
 // register the middleware
 app.use('/myCustomAPIEndpoint', Fetcher.middleware());
 
-app.use(function(req, res, next) {
+app.use(function (req, res, next) {
     // instantiated fetcher with access to req object
     const fetcher = new Fetcher({
         xhrPath: '/myCustomAPIEndpoint', // xhrPath will be ignored on the serverside fetcher instantiation
-        req: req
+        req: req,
     });
 
     // perform read call to get data
     fetcher
         .read('data_service')
-        .params({id: ###})
-        .end(function (err, data, meta) {
-        // handle err and/or data returned from data fetcher in this callback
+        .params({ id: 42 })
+        .then(({ data, meta }) => {
+            // handle data returned from data fetcher in this callback
+        })
+        .catch((err) => {
+            // handle error
         });
 });
 ```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -106,19 +106,22 @@ Fetcher.registerService(myDataService);
...
+        req: req,
     });
 
     // perform read call to get data
     fetcher
         .read('data_service')
-        .params({id: ###})
-        .end(function (err, data, meta) {
-        // handle err and/or data returned from data fetcher in this callback
+        .params({ id: 42 })
```
redonkulus, line 119: This is more of a style comment and less about the code changes, but I've never been a fan of this programming style. All of this code was written before async / await was available. I would prefer if the API were something simpler like:

```js
try { 
  // read(/* service */, /* params */, /* config */);
  const data = await fetcher.read('data_service', { id: 42 });
} catch (e) {
  console.log(e);
}
```

Maybe this is something we change for a 1.x release.

libs/util/httpRequest.js
```diff
@@ -9,154 +9,161 @@
...
         setTimeout(function () {
-            fn().then(resolve, reject);
+            _fetch.call(self).then(resolve, reject);
         }, delay);
     });
 }
 
-function io(options, controller) {
+function _fetch() {
+    var self = this;
```
redonkulus, line 65: nit: do we still need to use `self`?

libs/util/httpRequest.js
```diff
@@ -9,154 +9,161 @@
...
+    if (this._options.method === 'POST' && !this._options.retry.retryOnPost) {
         return false;
     }
 
-    return options.retry.statusCodes.indexOf(err.statusCode) !== -1;
+    return this._options.retry.statusCodes.indexOf(err.statusCode) !== -1;
 }
 
-function delayPromise(fn, delay) {
+// _retry is the onReject promise callback that we attach to the
```
redonkulus, line 28: Any thing to review in these changes or is it just reorganizing the code?

redonkulus (commented):
Overall this looks ok. I'm assuming there are not breaking changes here and just organization and code improvements?

README.md
pablopalacios: Yup, I agree. I only fixed the docs because there are people at my company requiring a more up to date docs. I've also revisited what we discussed on #275 and this is what I want to tackle next. I also think it's possible to make it backward compatible so we can adjust everything before going to v1 and remove all the other APIs.

libs/util/httpRequest.js
pablopalacios: We need it later in the fetch call itself. Most of the time we could use `this` and only self in the fetch promise. But having both would be annoying so I opted to only use one. I was also very tempted to use arrow functions so we wouldn't need this workaround. I was afraid that you would consider it a breaking change, so I kept the es5 style.

libs/util/httpRequest.js
pablopalacios: Not really, just reorganizing.

pablopalacios:
> Overall this looks ok. I'm assuming there are not breaking changes here and just organization and code improvements?

It's mainly refactoring. Calling `.end()` without a callback is still possible but I removed all the tests and added a warning for the users.

pablopalacios:
@redonkulus any blocker?

README.md
redonkulus: Makes sense to me. BC is ideal but not a blocker if it makes the API better and user friendly.

redonkulus (approved):

redonkulus:
Published https://github.com/yahoo/fetchr/releases/tag/v0.7.11

PR #522, opened 2024-08-15
fix: always return the same value when resolving requests multiple times

pablopalacios:

I found a bug in our system where we were resolving the request multiple times. The first time it was resolved it was fine (and so fetchr tests), but the second time it would resolve with an undefined value. The code was looking something like this:

```js
const request = fetchr.read('foo', { bar: 42 }); // request was triggered
track(request); // would call then internally
return request; // the consumer of this function would also call then
```

Fixing it was simple, we just had to keep the reference to fetch promise in `httpRequest` and always return `_request.then` in `FetchrHttpRequest.then` method.

However, that broke the abort support since we were not return an instance of `FetchrHttpRequest` anymore to the caller. To solve that, it was also necessary to store a reference to `FetchrHttpRequest` in the `Request` class in `fetchr.client.js`.

I added a bunch of more tests to cover this and other cases as well.

### DIFF
--- a/libs/fetcher.client.js
+++ b/libs/fetcher.client.js
@@ -38,6 +38,7 @@ function Request(operation, resource, options) {
     this._body = null;
     this._clientConfig = {};
     this._startTime = 0;
+    this._request = null;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


redonkulus (approved):
Interesting bug find, changes look good to me.

PR #523, opened 2024-08-18
feat: add support for async based resource operation handlers

pablopalacios:

This PR enables fetchr to be used without any callback.

## Before
```js
function create(req, resource, params, body, config, callback) { 
    callback(err, data, meta)
}
```

## After
```js
async function create({ req, resource, params, body, config }) {
    return { data, meta }
 }
```

It's still possible to use the old signature since we detect the new one based on the number or arguments defined in the handler.

This change also make it simple to create handlers that do not require any argument:

## Before
```js
function create(req, resource, params, body, config, callback) {
    callback({ ok: true });
}
```

## After
```js
async function create() {
    return { data: { ok: true } };
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ When an error occurs in your Fetchr CRUD method, you should throw an error objec
 ```js
 export default {
     resource: 'FooService',
-    read: async function create(req, resource, params, configs, callback) {
+    read: async function create(req, resource, params, configs) {
         const err = new Error('it failed');
         err.statusCode = 404;
         err.output = { message: 'Not found', more: 'meta data' };
--- a/README.md
+++ b/README.md
@@ -75,13 +75,13 @@ export default {
     // resource is required
     resource: 'data_service',
     // at least one of the CRUD methods is required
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -225,17 +225,17 @@ fetcher.updateOptions({
...
 ## Error Handling
 
-When an error occurs in your Fetchr CRUD method, you should return an error object to the callback. The error object should contain a `statusCode` (default 500) and `output` property that contains a JSON serializable object which will be sent to the client.
+When an error occurs in your Fetchr CRUD method, you should throw an error object. The error object should contain a `statusCode` (default 500) and `output` property that contains a JSON serializable object which will be sent to the client.
 
 ```js
 export default {
     resource: 'FooService',
-    read: function create(req, resource, params, configs, callback) {
+    read: async function create(req, resource, params, configs, callback) {
```
redonkulus, line 233:
```suggestion
    read: async function create(req, resource, params, configs) {
```

pablopalacios:
@redonkulus I tried with npm link on our main website code yes with both versions running (callback and async). All good!

PR #532, opened 2024-09-27
Fix send vulnerable to template injection that can lead to XSS

andrisecops:
passing untrusted user input - even after sanitizing it - to `SendStream.redirect()` may execute untrusted code

The vulnerability arises when untrusted user input is passed to the `SendStream.redirect()` function, even after sanitization. This can result in the execution of untrusted code. Successful exploitation of this vulnerability requires the following conditions: 1) The attacker must control the input to response.redirect(),  2) The express framework must not redirect before the template appears, and  3) The browser must not complete redirection before the user clicks on the link in the template. The impact of this vulnerability can lead to XSS attacks, compromising the security and integrity of the application.

CWE-79
CVE-2024-43799

## Patches
this issue is patched in `fetchr_send` 0.19.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1340,9 +1340,9 @@
       }
     },
     "node_modules/body-parser": {
-      "version": "1.20.2",
-      "resolved": "https://registry.npmjs.org/body-parser/-/body-parser-1.20.2.tgz",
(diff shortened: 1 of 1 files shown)


redonkulus:
Updated the package via dependabot PR

